PR #19, opened 2022-02-22
Profiles queries and messages implementation (desmos-std package)

bragaz:
This PR implements the Desmos `x/profiles` queries and messages, alongside useful mocks for contracts' developers tests.


packages/desmos-std/src/types.rs
```diff
@@ -1,13 +1,30 @@
...
+    limit: u64,
+    count_total: bool,
+    reverse: bool,
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Default, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub struct PageResponse {
+    next_key: Option<String>,
+    total: String,
```
dadamu, line 29:
```suggestion
     #[serde(default)]
    total: u64,
```
The total field type is u64. In addition, the total field would be skipped when total is equal to 0 so #[serde(default)] tag is needed.

packages/desmos-std/src/types.rs
```diff
@@ -1,13 +1,30 @@
...
 #[serde(rename_all = "snake_case")]
 pub enum DesmosRoute {
     Subspaces,
     Profiles,
 }
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub struct PageRequest {
+    key: Vec<u8>,
```
dadamu, line 18: Should we send back key as `String` to desmos app, as same as `PageResponse next_key` is `String`.

packages/desmos-std/src/msg_router.rs
```diff
@@ -0,0 +1,177 @@
...
+#[non_exhaustive] // missing chain-links and app-links messages (not necessary to me)
+pub enum DesmosMsg {
+    Profiles(ProfilesMsgs),
+}
+
+impl ProfilesMsgRouter<DesmosMsgRouter> for DesmosMsgRouter {
+    fn save_profile(
+        dtag: String,
+        creator: Addr,
+        nickname: Option<String>,
```
dadamu, line 43: I think Option should be remove here and checked by the the execute function which call `MsgRouter` api.

packages/desmos-std/src/msg_router.rs
```diff
@@ -0,0 +1,177 @@
...
+impl CustomMsg for DesmosMsgRouter {}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+#[non_exhaustive] // missing chain-links and app-links messages (not necessary to me)
+pub enum DesmosMsg {
+    Profiles(ProfilesMsgs),
+}
+
+impl ProfilesMsgRouter<DesmosMsgRouter> for DesmosMsgRouter {
```
dadamu, line 39: To build the `MsgRouter`, we should create empty `MsgRouter` firstly then rebuild it by the functions in the trait. The process is a bit strange to me.

I propose to have a `MsgRouterBuilder` to handle the `MsgRouter` creation, which is like as follows:

```rust
pub struct DesmosMsgRouterBuilder {}
impl DesmosMsgRouterBuilder{
    pub fn new() -> Self {
        DesmosMsgRouterBuilder{}
    }
}

impl ProfilesMsgRouterBuilder<DesmosMsgRouter> for DesmosMsgRouterBuilder {
    fn save_profile(...) -> DesmosMsgRouter {
        ...
    }
    ...
}
```

What do you think? @bragaz

packages/desmos-std/src/profiles/models_app_links.rs
```diff
@@ -0,0 +1,64 @@
...
+#[serde(rename_all = "snake_case")]
+pub enum AppLinkResult {
+    Success { value: String, signature: String },
+    Failed { error: String },
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub struct QueryApplicationLinksResponse {
+    pub links: Vec<ApplicationLink>,
```
dadamu, line 50: I found that desmos app returns `null` as value when array is empty so it should be `Option<Vec<ApplicationLink>>` to solve the null deserialize issue.

packages/desmos-std/src/profiles/models_app_links.rs
```diff
@@ -0,0 +1,64 @@
...
+    pub pagination: PageResponse,
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub struct QueryUserApplicationLinkResponse {
+    pub links: ApplicationLink,
```
dadamu, line 57:
```suggestion
    pub link: ApplicationLink,
```

packages/desmos-std/src/profiles/mocks.rs
```diff
@@ -0,0 +1,117 @@
...
+    models_blocks::UserBlock,
+    models_chain_links::{ChainConfig, ChainLink, ChainLinkAddr, Proof, Signature},
+    models_common::PubKey,
+    models_dtag_requests::DtagTransferRequest,
+    models_profile::{Account, Pictures, Profile},
+    models_relationships::Relationship,
+};
+use cosmwasm_std::{Addr, Binary, ContractResult};
+
+pub trait MockQueriesProfiles<T> {
```
dadamu, line 12: I think this trait could be moved to the src/query_mocks.rs with name `MockQueries`.

packages/desmos-std/src/profiles/mocks.rs
```diff
@@ -0,0 +1,117 @@
...
+    fn custom_query_execute(query: &T) -> ContractResult<Binary>;
+}
+
+/**
+This file contains some useful mocks of the Desmos x/profiles modules types ready made to be used
+in any test
+**/
+
+pub fn get_mock_profile() -> Profile {
```
dadamu, line 21: Should we wrap all mock functions into a struct?

packages/desmos-std/src/types.rs
bragaz: I tried to use `u64` as well, but it appears that the `PageResponse` JSON is Serialized as follow:
```json
"pagination": {
    "next_key": null,
    "total": "4"
  }
```
So the total is basically a `String`. This will make the deserialisation to u64 fail. Also when it's 0, it will be serialised as:
```json
"pagination": {
    "next_key": null,
    "total": "0"
  }
```

I've tested it locally and it happens to not work at all with `u64` unfortunately

packages/desmos-std/src/msg_router.rs
bragaz: I first did that thinking about the fact that those are optional fields but that not make much sense since they will later be saved as empty strings by the Desmos app.

packages/desmos-std/src/msg_router.rs
bragaz: @dadamu makes sense! 
I would call it `DesmosMsgBuilder`, since it builds messages and not the router itself.

packages/desmos-std/src/profiles/mocks.rs
bragaz: Yes, looks reasonable to do it. It will explicit it more clearly to devs when they need to use them

packages/desmos-std/src/profiles/models_app_links.rs
dadamu: In the rest api server it shows the right empty array, but the response to smart contract would be null instead of empty array.
The query subspaces test manually from my side, it returns null to the contract.
```json
{"subspaces":null,"pagination":{}}
```

dadamu: In the rest api server it shows the right empty array, but the response to smart contract would be null instead of empty array.
The query subspaces test manually from my side, it returns null to the contract.
```json
{"subspaces":null,"pagination":{}}
```

dadamu: @bragaz 
In the rest api server it shows the right empty array, but the response to smart contract would be null instead of empty array.
The query subspaces test manually from my side, it returns null to the contract.
```json
{"subspaces":null,"pagination":{}}
```

dadamu: You are right, I used default json marshal. Using cdc marshaling will get right result.

packages/desmos-std/src/types.rs
dadamu: You are right, I used default json marshal. Using cdc marshaling will get right result.

dadamu: @braze
Finally I found a better solution which is using cosmwasm_std::Uint64 and cosmwasm_std::Binary.
The pagination could be:
```rust
pub struct PageResponse {
    next_key: Option<Binary>,
    total: Uint64,
}
```

packages/desmos-std/src/profiles/msg_router.rs
```diff
@@ -0,0 +1,34 @@
+use cosmwasm_std::{Addr, CosmosMsg};
+
+pub trait ProfilesMsgRouter<T> {
```
dadamu, line 3: I would name it ProfilesMsgBuilder.

packages/desmos-std/src/profiles/query_routes.rs
```diff
@@ -0,0 +1,47 @@
...
+    Profile {
+        user: Addr,
+    },
+    IncomingDtagTransferRequests {
+        receiver: Addr,
+        pagination: Option<PageRequest>,
+    },
+    Relationships {
+        user: Addr,
+        subspace_id: u64,
```
dadamu, line 18: It should be cosmwasm_std::Uint64 instead. The u64 response from desmos is string type.

packages/desmos-std/src/types.rs
dadamu: @braze Finally I found a better solution which is using cosmwasm_std::Uint64 and cosmwasm_std::Binary.
The pagination stuff could be:
```rust
pub struct PageRequest {
    key: Option<Binary>,
    offset: Uint64,
    limit: Uint64,
    count_total: bool,
    reverse: bool,
}

pub struct PageResponse {
    next_key: Option<Binary>,
    total: Uint64,
}
```

bragaz: It worked fine 👍 Great Job!

packages/desmos-std/src/profiles/querier.rs
```diff
@@ -0,0 +1,62 @@
...
+        models_chain_links::{QueryChainLinksResponse, QueryUserChainLinkResponse},
+        models_dtag_requests::QueryIncomingDtagTransferRequestResponse,
+        models_profile::QueryProfileResponse,
+        models_relationships::QueryRelationshipsResponse,
+    },
+    types::PageRequest,
+};
+use cosmwasm_std::{Addr, StdResult, Uint64};
+
+pub trait ProfilesQuerier {
```
dadamu, line 17: Could we build querier for each modules individually? I think we can create querier from the `cosmos_std::Deps` instead of creating our `Deps` with fixed query type.
Just tried it with the following branch:
https://github.com/desmos-labs/desmos-contracts/blob/paul/subspaces-queries-clean/packages/desmos-std/src/subspaces/querier.rs#L18

bragaz: As long as it simplifies devs' works it's fine. Plus, it will let us to better organise code base on each module independently from one module to another

bragaz:
@dadamu I've took a look at your branch just to align mine in term of module refactoring.
I think everything now is pretty much ok, I've fixed some `Uint64` types and rollbacked them to `u64` since they couldn't be properly deserialised by the desmos app. I've also tested a contract and it worked fine. 

I only have some concerns about having multiple msg router for each module. For instance, if someone need more than one module's messages,  that means that they will need to build more than 1 message builder, which could lead to a greater gas consumption. What do you think about this? @dadamu

dadamu:
@bragaz Nice catch on u64 stuff!

About the msg routers, you are right, it could lead to a greater gas consumption when importing multiple modules. In the other hand, the contract which uses only one module doesn't need to import all the modules in desmos. To be honest, I am not sure how much gas would increase. But I guess it worth to do it to have more explicit to each modules, and the gas increasing would be not much.

bragaz:
@dadamu, Yes I'm not sure about how much the gas increase would be, I guess very little btw. We can proceed this way as it let us to have a better separation among modules. I will open an issue to write some tests about this

packages/desmos-std/src/profiles/mocks.rs
```diff
@@ -0,0 +1,408 @@
...
+                    .to_string(),
+            },
+            creation_time: "2022-02-21T13:18:57.800827Z".to_string(),
+        }
+    }
+}
+
+/// Replacement for cosmwasm_std::testing::mock_dependencies
+/// this use our CustomQuerier to use desmos querier
+pub fn mock_dependencies_with_custom_querier(
```
dadamu, line 130: I am trying about if we can mock all the modules' `QueryRouter` to one `deps`.
Currently it seems impossible since the generic T of `MockQuerier<T>` does not accept trait as parameter, so the `deps` only can be mocked for each modules individually, as a result, the following code cannot be tested:
```rust
fn query(deps: Deps) -> StdResult<QueryResponse> {
    let subspaces_querier = SubspacesQuerier::new(deps.querier.deref());
    let subspace = subspaces_querier.query_subspace(1)
    let profiles_querier = ProfilesQuerier::new(deps.querier.deref());
    let profile = profiles_querier.query_profile("user");
    QueryResponse{
       subspace, profile
   }
}
```
Should we have a general `DesmosQueryRouter` to solve the problem? Or just require developer only use one querier in one function to let the function be testable.
Any idea of it @bragaz ?

packages/desmos-std/src/profiles/msg_router.rs
```diff
@@ -0,0 +1,189 @@
...
+    }
+}
+
+impl Default for ProfilesMsgBuilder {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl ProfilesMsgBuilder {
```
dadamu, line 35: Should we test the msg builder?

packages/desmos-std/src/profiles/msg_routes.rs
```diff
@@ -0,0 +1,57 @@
+use cosmwasm_std::Addr;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum ProfilesMsgs {
```
dadamu, line 7: I think it lost `AppLink` and `ChainLink`.

packages/desmos-std/src/profiles/mocks.rs
```diff
@@ -0,0 +1,408 @@
...
+        });
+    OwnedDeps::<_, _, _, ProfilesQueryRouter> {
+        storage: MockStorage::default(),
+        api: MockApi::default(),
+        querier: custom_querier,
+        custom_query_type: PhantomData,
+    }
+}
+
+pub trait MockQueries<T> {
```
dadamu, line 146: Since the instance which execute `query` function should be `Querier` instead of `QueryRouter`, it should be replaced by a `MockProfilesQuerier` like:
```rust
pub struct MockProfilesQuerier {}

impl MockSubspacesQuerier {
    fn custom_query_execute(query: &ProfilesQueryRouter) ->  ContractResult<Binary> { .. }
}
```

packages/desmos-std/src/profiles/msg_routes.rs
bragaz: I think there's no point to add those messages right know since it will not be that straightforward to implement them at the moment. Since the contract is the one who signs the messages, users cannot create links easily. I suggest we keep the #[non_exhaustive] notation till we implement them in the future

dadamu: I think it is useful to create link for the contract which is the owner of subspace. I believe that it is impossible to create the link for user in the contract since the signer is contract as you said. I am okay to keep #[non_exhaustive] notation.

packages/desmos-std/src/profiles/mocks.rs
bragaz: I guess it would be better to only have one generic `DesmosQueryRouter` with a generic `QueryRoute` enum that contains all the nested desmos queries like we did at first:

```rust
#[serde(rename_all = "snake_case")]
pub struct DesmosQueryRouter {
    pub route: DesmosRoute,
    pub query_data: QueryRoute ,
}

pub enum QueryRoute {
    Profiles(ProfilesQueryRoute)
    Subspaces(SubspacesQueryRoute)
}
```

We shouldn't limit the possibility to make more than one custom query per function IMO.

packages/desmos-std/src/profiles/msg_router.rs
bragaz: Why not, it's trivial to test them and it's a good way to check unwanted changes

packages/desmos-std/src/profiles/msg_routes.rs
bragaz: @dadamu I didn't understand what do you mean by:

> the contract which is the owner of subspace. 

Do you mean that it would be good for contracts to be able to create links in the case they are set as subspace's owners/admins?

dadamu: @bragaz Yes, that is. Contract can be the subspace owner as well. Then everything about subspace can be controlled by the contract.

packages/desmos-std/src/profiles/mocks.rs
dadamu: @bragaz I agree with the idea. It is also easy to be testable as well.

dadamu:
@bragaz Rethinking about the reason why we can't parse `Uint64` properly, I found it is because that desmos applies the default `json.Unmarshal` to parse the request message.
https://github.com/desmos-labs/desmos/blob/leonardo/contracts-tests/x/profiles/wasm/querier.go#L33

To let the request as similar as the rpc json format, we can parse the request message with the codec way, like to make sure `Uin64` parsed properly:
https://github.com/desmos-labs/desmos/blob/paul/wasm-uint64-parse/x/subspaces/wasm/querier.go#L43

packages/desmos-std/src/mock.rs
```diff
@@ -1,163 +1,195 @@
...
-            attachments: Some(vec![]),
-            poll: Some(Poll {
-                question: "".to_string(),
-                provided_answers: vec![],
-                end_date: "".to_string(),
-                allows_multiple_answers: false,
-                allows_answer_edits: false,
-            }),
-            creator: String::from("default_creator"),
+    fn test_query_profile() {
```
dadamu, line 50: In the general `mock` lib, the thing only should be tested is if the binding of `mock dependency` to the module mock querier works properly or not.
We can remain only one test function to test the binding, then move the rest of them to the module `mock lib`.

dadamu:
@bragaz Ah, I mean that if we apply codec as parser, the types of all request would be `json.RawMessage`, and all parsed by codec.

packages/desmos-std/src/mock.rs
bragaz: @dadamu Thanks for pointing this out. I'm moving the querier tests inside `querier.rs` and add some proper tests here

packages/desmos-std/src/profiles/models_app_links.rs
```diff
@@ -0,0 +1,51 @@
...
+#[serde(rename_all = "snake_case")]
+pub struct Data {
+    pub application: String,
+    pub username: String,
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub struct TimeoutHeight {
+    pub revision_number: u64,
```
dadamu, line 26: After applying codec in the desmos side, we can replace all `u64` into `Uint64` safely.

packages/desmos-std/src/profiles/mod.rs
```diff
@@ -0,0 +1,13 @@
...
+pub mod models_chain_links;
+pub mod models_common;
+pub mod models_dtag_requests;
+pub mod models_profile;
+pub mod models_query;
+pub mod models_relationships;
+pub mod msg;
+pub mod msg_builder;
+pub mod querier;
+pub mod query_router;
```
dadamu, line 13: I would naime it query only.

packages/desmos-std/src/types.rs
```diff
@@ -1,13 +1,27 @@
...
```
dadamu, line 15: It could be `Option<Binary>` after applying codec.

packages/desmos-std/src/profiles/mod.rs
bragaz: which one? The query_router?

packages/desmos-std/src/profiles/models_profile.rs
```diff
@@ -0,0 +1,59 @@
...
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub struct Account {
+    #[serde(rename = "@type")]
+    pub proto_type: String,
+    pub address: Addr,
+    pub pub_key: PubKey,
+    pub account_number: String,
+    pub sequence: String,
```
dadamu, line 45:
```suggestion
    pub account_number: Uint64,
    pub sequence: Uint64,
```
```suggestion
    pub account_number: String,
    pub sequence: String,
```

dadamu, line 45: These two should be `Uint64` as well.

dadamu, line 45: These two should be `Uint64` as well.

packages/desmos-std/src/profiles/models_common.rs
```diff
@@ -0,0 +1,10 @@
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub struct PubKey {
+    #[serde(rename = "@type")]
+    pub proto_type: String,
+    pub key: String,
```
dadamu, line 9:
```suggestion
    pub key: Binary,
```

The key from desmos chain is base64 encoded. It could be Binary type.

packages/desmos-std/src/profiles/msg.rs
```diff
@@ -0,0 +1,74 @@
...
+    LinkApplication {
+        sender: Addr,
+        link_data: Data,
+        call_data: String,
+        source_port: String,
+        source_channel: String,
+        timeout_height: TimeoutHeight,
+        timeout_timestamp: Uint64,
+    },
+    CreateRelationship {
```
dadamu, line 53: I think we can remove relationships since it doesn't belong to `x/profiles` anymore.

dadamu (changes requested):
Overall looks good to me! Just a couple things could be improved.

packages/desmos-std/src/profiles/msg.rs
bragaz: I would like to handle the split of modules in a different PR

dadamu (approved):
Ready to go from my side 🚀

PR #23, opened 2022-03-14
Improve `DesmosQuery` and `DesmosMsg` with serde tag

### LINKED ISSUE #21: Improve `DesmosQuery` and `DesmosMsg` with serde tag
Currently the structure of `DesmosMsg` is like:

```rust
#[serde(rename_all = "snake_case")]
pub struct DesmosMsg {
    pub route: DesmosRoute,
    pub msg_data: DesmosMsgRoute,
}
```

It could be simplified with serde tag attribute, say:
```rust
#[serde(rename_all = "snake_case", tag = "route", content = "msg_data")]
pub enum DesmosMsg {
    Profiles(ProfilesMsg),
    Subspaces(SubspacesMsg),
}
```
To produce the json like:
```json
{
    "route": "subspace",
    "msg_data": {
        "create_subspace": {
            "name": "test",
            "description": "test",
            "......"
        }
    }
}
```

We can do the same way to `DesmosQuery` as well.
It will reduce some code related to `DesmosRoute`. What do you think? @bragaz 

### Note:

it will break the current wasm part of modules in the desmos chain since the json format is different from the current one.

Reference: https://serde.rs/enum-representations.html#adjacently-tagged

dadamu:
Close #21 

This PR applies the serde tag to indicate the route instead of `DesmosRoute` object. In addition, `DesmosRoute` is removed since it is never used with this change.


packages/desmos-std/src/msg.rs
```diff
@@ -2,18 +2,20 @@ use cosmwasm_std::{CosmosMsg, CustomMsg};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
-use crate::{profiles::msg::ProfilesMsg, subspaces::msg::SubspacesMsg, types::DesmosRoute};
+use crate::{profiles::msg::ProfilesMsg, subspaces::msg::SubspacesMsg};
 
+// Use serde tag in order to produce the certain json format
```
bragaz, line 7:
```suggestion
// Use the serde `rename_all` tag in order to produce the following json file structure
```

packages/desmos-std/src/query.rs
```diff
@@ -2,49 +2,42 @@ use cosmwasm_std::CustomQuery;
...
-pub struct DesmosQuery {
-    pub route: DesmosRoute,
-    pub query_data: DesmosQueryRoute,
-}
-
-impl CustomQuery for DesmosQuery {}
-
+use crate::{profiles::query::ProfilesQuery, subspaces::query::SubspacesQuery};
+
+// Use serde tag in order to produce the certain json format
```
bragaz, line 7:
```suggestion
// Use the serde `rename_all` tag in order to produce the following json file structure
```

bragaz (approved):
Looks ready 👍

bragaz:
Closes #21

PR #26, opened 2022-03-14
Enable conditional compilation (features)

### LINKED ISSUE #24: Add conditional compilation to desmos-std
In order to optimise the usage of desmos-std by devs in their contract, we can add [conditional compilation](https://doc.rust-lang.org/reference/conditional-compilation.html) to its packages. Conditional compilation allows to build only the features that are needed in a specific context.   
For instance, if someone wants to write a contract which only use `subspaces` messages, it's useless for them to also include other modules. Including them all will result in:  
- an increased contract size;
- a higher deployment cost.

To solve these problems, conditional compilation come useful, we just need to:
1) Updated the `config.toml`:

```toml
...
[features]
profiles = []
subspaces = []
...
```

2) Update the code where needed, with macros for features:
```rust
#[serde(rename_all = "snake_case", tag = "route", content = "msg_data")]
pub enum DesmosMsg {
    #[cfg(feature = "profiles")]
    Profiles(ProfilesMsg),
    #[cfg(feature = "subspaces")]
    Subspaces(SubspacesMsg),
}
```

Note:
Special thanks to @manu0466 to bring this to my attention

@dadamu what do you think about this?

bragaz:
This PR introduce conditional compilation as described here: #24.
Closes #24.

### DIFF
--- a/packages/desmos-std/Cargo.toml
+++ b/packages/desmos-std/Cargo.toml
@@ -21,3 +21,6 @@ cosmwasm-schema = { version = "1.0.0-beta" }
 profiles = []
 subspaces = []
 relationships = []
+query = []
+msg = []
+mocks = []
--- a/packages/desmos-std/Cargo.toml
+++ b/packages/desmos-std/Cargo.toml
@@ -18,6 +18,7 @@ serde = { version = "1.0.103", default-features = false, features = ["derive"] }
 cosmwasm-schema = { version = "1.0.0-beta" }
 
 [features]
+default = ["profiles", "subspaces", "relationships", "query", "msg"]
 profiles = []
 subspaces = []
 relationships = []
--- a/packages/desmos-std/Cargo.toml
+++ b/packages/desmos-std/Cargo.toml
@@ -16,3 +16,8 @@ serde = { version = "1.0.103", default-features = false, features = ["derive"] }
 
 [dev-dependencies]
 cosmwasm-schema = { version = "1.0.0-beta" }
+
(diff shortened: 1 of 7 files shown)


packages/desmos-std/src/msg.rs
```diff
@@ -28,35 +41,67 @@ impl Into<CosmosMsg<DesmosMsg>> for DesmosMsg {
...
+        Self::Relationships(msg)
+    }
+}
+
 #[cfg(test)]
 mod tests {
-    use crate::{msg::DesmosMsg, profiles::msg::ProfilesMsg, subspaces::msg::SubspacesMsg};
+    use crate::msg::DesmosMsg;
+
+    #[cfg(feature = "profiles")]
```
dadamu, line 69: I think the feature tag could be removed in the test since the everything should be tested.

packages/desmos-std/Cargo.toml
```diff
@@ -16,3 +16,11 @@ serde = { version = "1.0.103", default-features = false, features = ["derive"] }
 
 [dev-dependencies]
 cosmwasm-schema = { version = "1.0.0-beta" }
+
+[features]
```
dadamu, line 20: It would be useful to have default feature.
https://doc.rust-lang.org/cargo/reference/features.html#the-default-feature

bragaz: @dadamu added them

dadamu (approved):

PR #29, opened 2022-03-18
Features compilation fix

manu0466:
This PR fixes some compilation errors when compiling `desmos-std` without the default features and adds some CI tests to prevent regressions


packages/desmos-std/src/relationships/mod.rs
```diff
@@ -1,9 +1,11 @@
-#[cfg(not(target_arch = "wasm32"))]
+#[cfg(all(not(target_arch = "wasm32"), feature = "query"))]
```
dadamu, line 1:
```suggestion
#[cfg(all(not(target_arch = "wasm32"), feature = "mocks"))]
```

.github/workflows/desmos_std.yml
```diff
@@ -75,3 +75,58 @@ jobs:
...
+
+      - name: Build feature (relationships)
+        run: cargo build --no-default-features --features relationships
+        working-directory: ${{env.working-directory}}
+
+      - name: Build feature (relationships+query)
+        run: cargo build --no-default-features --features relationships,query
+        working-directory: ${{env.working-directory}}
+
+      - name: Build feature (msg)
```
dadamu, line 126: Adding build feature for each modules msg would be useful, say:

```yml
- name: Build feature (profiles+msg)
  run: cargo build --no-default-features --features profiles,msg
  working-directory: ${{env.working-directory}}
```

.codecov.yml
```diff
@@ -39,3 +39,4 @@ ignore:
   - "src/lib.rs"
   - "src/types.rs"
   - "src/state.rs"
+  - "packages/desmos-std/**/mock*.rs"
```
dadamu, line 42: Is there any reason to ignore it?

manu0466: Since are mocks I don't think is actually useful to include that code in the codecov report

bragaz: I think it's safe to ignore them since they're only used for testing

bragaz (commented):
After fixing @dadamu requests it should be good to be merged

.codecov.yml
dadamu: Okay, got it. 👍

bragaz (approved):
Good to go! 🚀

dadamu (approved):
🚀

PR #30, opened 2022-03-18
Iterator to better manage data query

### LINKED ISSUE #31: Iterator for data query
The idea is to give to the developers an easier way to query and perform operations with the data present on the chain leveraging the standard `Iterator` crate of Rust.

Using our current implementation if a developer wants to operate with the app-links need to do something like this:
```
let profiles_querier = ProfilesQuerier::new(...);
let mut data_available = true;
let mut current_offset: u64 = 0;
while data_available {
    let response = profiles_querier.query_application_links(
       Some(address)), None, None, 
       Some(PageRequest {
           key: None,
           offset: Uint64::from(current_offset),
           limit: Uint64::from(10),
           count_total: false,
           reverse: false,
       }))?;
       if response.links.is_empty() {
           data_available = false
       } else { 
          // Scan app links and perform operations
          current_offset += response.links.len() as u64;
       }
}
``` 

With an iterator the same code can be something like this:
```
let profiles_querier = ProfilesQuerier::new(...);
let it = profiles_querirer.query_application_links_it(Some(address), None, None, 10);
for app_link_result in it {
    let app_link = app_link_result?;
    // Do operations with the fetched app link...
}

### PR DESCRIPTION
**Pull request by manu0466**
Closes #31.

# Drawbacks
Since the elements fetched from the chain are wrapped arround a `StdResult` the elements are being copied into it, this limitation can be overcommed exposing an `Iterator` that simply abstract away the page fetching operations exposing a sequence of pages insted a sequence of elements.

### CODE REVIEW

**Comment by dadamu** [comment]
Using iterator pattern is awesome, but the current implementation is using offset to get the next page. Using next_key to paginate  would be better than offset pagination.

[Reference](https://github.com/cosmos/cosmos-sdk/blob/23e864bc987e61af84763d9a3e531707f9dfbc84/proto/cosmos/base/query/v1beta1/pagination.proto#L14)

What do you think @manu0466 @bragaz

**Comment by manu0466** [comment]
Thanks to point out that @dadamu, I will update the iterator to use the `next_key` field.

File: packages/desmos-std/src/iter/page_iterator.rs [hunk 1]
```diff
@@ -0,0 +1,211 @@
...
+        assert!(item.unwrap().is_err());
+
+        // Since an error occurred when fetching the page the second item should be None
+        // to signal that the iterator don't have more items.
+        let item = it.next();
+        assert!(item.is_none());
+    }
+
+    #[test]
+    fn test_iterations_with_partial_page() {
```
dadamu, line 186: Could you and the test with the case having next page?

manu0466: A test where the first page is full and the second one is partial?

dadamu: @manu0466 yeah, it is.

manu0466: Test added

bragaz (approved):

dadamu (approved):

PR #32, opened 2022-03-23
Moved the desmos-std package to the new desmos-bindings repo

### LINKED ISSUE #28: Move desmos-std in its own repository
I think that to pursue maintainability in the future, we should move out desmos-std in its own repository and then import it in all the future smart-contract that we will implement. What do you think @dadamu?

manu0466:
This PR fixes #28.
- Moved the `desmos-std` package to the new [desmos-bindings](https://github.com/desmos-labs/desmos-bindings) repo;
- Updated the CI to skip some jobs since we don't have at the moment a package that actualy builds in the repo.


contracts/example/Cargo.toml
```diff
@@ -0,0 +1,53 @@
+[package]
+name = "example"
+version = "0.1.0"
+authors = ["Manuel Turetta <manuel.turetta94@gmail.com>"]
+edition = "2018"
+
+exclude = [
+  # Those files are rust-optimizer artifacts. You might want to commit them for convenience but they should not be part of the source code publication.
```
bragaz, line 8:
```suggestion
  # These files are rust-optimizer artifacts. You might want to commit them for convenience but they should not be part of the source code publication.
```

bragaz (approved):
👍

dadamu (approved):

PR #34, opened 2022-07-11
docs(adr): POAP Manager Contract

bragaz:


docs/architecture/adr-002-event-manager-contract.md
```diff
@@ -0,0 +1,95 @@
...
+```
+
+* The `admin` identifies the user that controls the contract;
+* The `poap_contract` that the event manager control;
+* The `subspace_id` identifies the dApp where the contract lives;
+
+#### Execute
+```rust
+pub enum ExecuteMsg{
+  TryMint{},
```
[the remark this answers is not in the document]
bragaz: I did not add the `subspace_id` field to the message because it's part of the contract state and initialised with the contract `InstatiateMsg`. This to force the checks on a particular subspace to avoid users looking for created posts and comments and likes in other subspaces. Why this? Because my original idea was to check if a user has made any posts, reactions, comment and let them mint the poap after that checks.
Sadly, Desmos still not support the query of posts by users, so that would be an additional work for the contract and I think it will increase the costs of computation quite a lot. Looking at your suggestion I thought that, to simplify this, we can make them comment and react only 1 specific post (so we can pass the id to chain). The problem will still subsist if we require them also to create a post but, since a comment is a post, I think we can rely only on those two checks.

docs/architecture/adr-002-event-manager-contract.md
```diff
@@ -0,0 +1,96 @@
...
+## Specifications
+Here below the specifications for the contract's messages:
+
+### Messages
+
+#### Instantiate
+```rust
+pub struct InstantiateMsg {
+  pub admin: Addr,
+  pub poap_instantiate_msg: POAPContractInstantiateMsg,
```
dadamu, line 40: It is missing POAP contract address.

docs/architecture/adr-002-event-manager-contract.md
```diff
@@ -0,0 +1,96 @@
...
+of the contract.
+
+### Query
+```rust
+pub enum QueryMsg {
+  /// Return a ManagerInfoResponse containing the useful information of the Manager contract
+  ManagerInfo{},
+
+  /// Inherited from the POAP Contract
+  EventInfo{}
```
dadamu, line 78: `EventInfo` can be queried by POAP contract.  We should show the target POAP contract address, instead.

docs/architecture/adr-002-event-manager-contract.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Query
+```rust
+pub enum QueryMsg {
+  /// Return a ConfigResponse containing the configuration info of the Manager contract
+  Config{},
+}
+```
+
+#### ManagerInfo
```
dadamu, line 82: It should be renamed to `Config`

docs/architecture/adr-002-event-manager-contract.md
```diff
@@ -0,0 +1,94 @@
...
+* The `poap_contract_code_id` identifies the code id of the contract necessary for the instantiation;
+* The `poap_instantiate_msg` instantiate the poap contract that the manager controls;
+* The `subspace_id` identifies the dApp where the contract lives;
+* THe `cosmoverse_post_id` identifies the cosmoverse post to wich the user will react.
+
+#### Execute
+```rust
+pub enum ExecuteMsg{
+  Claim{post_id: u64},
+  MintTo{user: String},
```
dadamu, line 58:
```suggestion
  MintTo{user: Addr},
```

docs/architecture/adr-002-event-manager-contract.md
```diff
@@ -0,0 +1,94 @@
...
+  UpdateAdmin{new_admin: String}
```
dadamu, line 59:
```suggestion
  UpdateAdmin{new_admin: Addr}
```

docs/architecture/adr-002-event-manager-contract.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Specifications
+Here below the specifications for the contract's messages:
+
+### Messages
+
+#### Instantiate
+```rust
+pub struct InstantiateMsg {
+  pub admin: Addr,
```
dadamu, line 40:
```suggestion
  pub admin: String,
```

dadamu: @bragaz  InstantiateMsg admin address is not verified. So it should be `String`.

dadamu (approved):
Great to go!

PR #35, opened 2022-07-19
feat: POAP contract implementation

manu0466:
This PR provide an implementation of the POAP contract as described in [ADR-001](https://github.com/desmos-labs/desmos-contracts/blob/master/docs/architecture/adr-001-poap-contract.md).


contracts/poap/src/contract_tests.rs
```diff
@@ -0,0 +1,827 @@
...
+        };
+        app.execute_contract(
+            Addr::unchecked(MINTER),
+            poap_contract_addr.clone(),
+            &msg,
+            &vec![],
+        )
+        .unwrap();
+
+        // Test that minter can call mint to
```
bragaz, line 663:
```suggestion
        // Test that admin can call mint to
```

contracts/poap/src/contract_tests.rs
```diff
@@ -0,0 +1,827 @@
...
+        // Admin should not be authorized to update the event info
+        assert!(result.is_err());
+
+        let result = app.execute_contract(
+            Addr::unchecked(CREATOR),
+            poap_contract_addr.clone(),
+            &msg,
+            &vec![],
+        );
+        // Admin should not be authorized to update the event info
```
bragaz, line 709:
```suggestion
        // Creator is authorised to update the event info
```

contracts/poap/src/contract_tests.rs
```diff
@@ -0,0 +1,827 @@
...
+        assert!(result.is_err());
+
+        let result = app.execute_contract(
+            Addr::unchecked(ADMIN),
+            poap_contract_addr.clone(),
+            &msg,
+            &vec![],
+        );
+        // Admin can update minter
+        assert!(result.is_ok());
```
bragaz, line 952: Can we add also one where the `MINTER` tries to update the minter?

contracts/poap/src/msg.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub creator: String,
+    /// Time at which the event begins.
+    pub start_time: Timestamp,
+    /// Time at which the event ends.
+    pub end_time: Timestamp,
+    /// Max amount of poap that a single user can mint.
+    pub per_address_limit: u32,
+    /// Identifies a valid IPFS URI corresponding to where the assets and metadata of the POAPs are stored.
+    pub base_poap_uri: String,
+    /// Uri
```
bragaz, line 32:
```suggestion
    /// Uri of the poap event
```

contracts/poap/src/msg.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub event_uri: String,
+    pub cw721_code_id: u64,
```
bragaz, line 34:
```suggestion
    /// Id of the cw721 contract that the POAP contract initialise.
    pub cw721_code_id: u64,
```

contracts/poap/src/msg.rs
```diff
@@ -0,0 +1,100 @@
...
+/// Response to [`QueryMsg::EventInfo`].
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct QueryEventInfoResponse {
+    /// Address of who created the event.
+    pub creator: Addr,
+    /// Time at which the event starts.
+    pub start_time: Timestamp,
+    /// Time at witch the event ends.
+    pub end_time: Timestamp,
+    /// IPFS uri where is stored the event's metadata.
```
bragaz, line 98:
```suggestion
    /// IPFS uri where the event's metadata are stored
```

contracts/poap/src/msg.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+/// Response to [`QueryMsg::EventInfo`].
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct QueryEventInfoResponse {
+    /// Address of who created the event.
+    pub creator: Addr,
+    /// Time at which the event starts.
+    pub start_time: Timestamp,
+    /// Time at witch the event ends.
```
bragaz, line 96:
```suggestion
    /// Time at which the event ends.
```

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,451 @@
...
+const CONTRACT_NAME: &str = "crates.io:poap";
+const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
+// actions consts
+const ACTION_ENABLE_MINT: &str = "enable mint";
+const ACTION_DISABLE_MINT: &str = "disable mint";
+const ACTION_MINT: &str = "mint";
+const ACTION_MINT_TO: &str = "mint to";
+const ACTION_UPDATE_EVENT_INFO: &str = "update event info";
+const ACTION_UPDATE_ADMIN: &str = "update admin";
+const ACTION_UPDATE_MINTER: &str = "update minter";
```
dadamu, line 31: I would have a file `action.rs` to store them. And make the value in `snake case` .

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,451 @@
...
+                minter: env.contract.address.to_string(),
+            })?,
+            funds: info.funds,
+            label: "poap cw721".to_string(),
+        }
+        .into(),
+        id: INSTANTIATE_CW721_REPLY_ID,
+        gas_limit: None,
+        reply_on: ReplyOn::Success,
+    }];
```
dadamu, line 133: It can be simplified to:
```rust
SubMsg::reply_on_success(
    wasm_instantiate(msg.cw721_code_id, &msg.cw721_initiate_msg, info.funds, "poap_cw721".into())?,
    INSTANTIATE_CW721_REPLY_ID,
)
```

Reference: 
[submessage](https://github.com/CosmWasm/cosmwasm/blob/main/packages/std/src/results/submessages.rs#L55)
[wasm_instantiate](https://github.com/CosmWasm/cosmwasm/blob/9bf920b00b76f31f3c73ed3632cd07cbd6a94b7b/packages/std/src/results/cosmos_msg.rs#L201)

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,451 @@
...
+        token_uri: Some(event_info.base_poap_uri),
+        extension: event_info.event_uri,
+    });
+
+    let cw721_address = CW721_ADDRESS.load(deps.storage)?;
+    let response_msg = CosmosMsg::Wasm(WasmMsg::Execute {
+        contract_addr: cw721_address.to_string(),
+        msg: to_binary(&mint_msg)?,
+        funds: vec![],
+    });
```
dadamu, line 272: It can be simplified to:
```rust
wasm_execute(
            cw721_address,
            &mint_msg,
            info.funds,
        )
```

Reference: [wasm_execute](https://github.com/CosmWasm/cosmwasm/blob/9bf920b00b76f31f3c73ed3632cd07cbd6a94b7b/packages/std/src/results/cosmos_msg.rs#L218)

contracts/poap/src/state.rs
```diff
@@ -0,0 +1,123 @@
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+use cosmwasm_std::{Addr, Timestamp};
+use cw_storage_plus::{Item, Map};
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct Config {
+    pub admin: Addr,
```
dadamu, line 9: I think we can use [`cw_controllers::Admin`](https://github.com/CosmWasm/cw-plus/blob/main/packages/controllers/src/admin.rs#L29) to handle the admin state.
What do you think?

contracts/poap/src/state.rs
```diff
@@ -0,0 +1,123 @@
...
+    pub minter: Addr,
```
dadamu, line 10: As well as minter, it is a kind of admin.

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,451 @@
...
+        });
+    }
+
+    // Check if the event is ended
+    if event_info.is_ended(&env.block.time) {
+        return Err(ContractError::EventTerminated {
+            current_time: env.block.time,
+            end_time: event_info.end_time,
+        });
+    }
```
dadamu, line 232: Can we merge them into one method `event_info.validate_date()`?

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,451 @@
...
+}
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn execute(
+    deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+    msg: ExecuteMsg,
+) -> Result<Response, ContractError> {
+    match msg {
```
dadamu, line 159: Should we have a function `msg.validate()` before executing?  What do you think @manu0466, @bragaz?

contracts/poap/src/msg.rs
```diff
@@ -0,0 +1,101 @@
...
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct InstantiateMsg {
+    /// Address of who will have the right to administer the contract.
+    pub admin: String,
+    /// Address of who can call the [`ExecuteMsg::MintTo`] other then the admin.
+    pub minter: String,
+    /// Id of the CW721 contract to initialize together with this contract.
+    pub cw721_code_id: u64,
```
dadamu, line 13: Should we use `Uint64` as input and output instead? The normal JSON `u64` is represented as a `number`, but Amino JSON for cosmos ecosystem `u64` is represented as `string`. What do you think  @bragaz, @manu0466

manu0466: Yes, even if the signing with ledger is working with u64 it's better to use `Uint64` since can happen that a Javascript client does not represents correctly integer with more then 53 bits

contracts/poap/src/contract.rs
manu0466: Yes snake case is better so that they have the same format of `ExecuteMsg`!
Regarding moving to a different file I think it's a bit over separation, no one beside the varius execution functions is using this consts and even if some one else will want to extend this contract probably don't need this strings.

contracts/poap/src/state.rs
manu0466: That struct stores the admin as `Option<Addr>`, in our case the admin address is always defined and using that will just add boilerplate code to unwrap the address.

contracts/poap/src/contract.rs
manu0466: Thinking about having the validation logic in a single place seems reasonable to me, we can have this methd that checks that the received data is correct but i prefer to have the permissions checks inside the execution function to make explicit that the authorization checks where properly performed.

contracts/poap/src/contract.rs
manu0466: With this we will move some contract logic inside `EventInfo` i do prefer keeping the responsibilities separated between them

contracts/poap/src/contract.rs
dadamu: Yeah, permissions checks only can be performed inside the execution process, `msg.validate` is for checking if the value of msg is valid, say, the check of `end_time` must be after `start_time` inside the msg can be validated here.
With this function, the execute function can be focused on verifying `permission` or `state` except for the msg data.

contracts/poap/src/state.rs
```diff
@@ -0,0 +1,123 @@
...
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct EventInfo {
+    pub creator: Addr,
+    pub start_time: Timestamp,
+    pub end_time: Timestamp,
+    pub base_poap_uri: String,
+    pub event_uri: String,
+}
+
+pub const CONFIG: Item<Config> = Item::new("config");
```
dadamu, line 25: Can we merge `CW721_ADDRESS` into `Config`? It would be useful to show the contract address.

manu0466: Actualy the cw721 address is returned in `QueryConfigResponse`,  I can mege CW721_ADDRESS and CONFIG btw.

dadamu: Oh, yeah. Just consider if let `QueryConfigResponse` returns config instead seperately fields.

contracts/poap/src/msg.rs
```diff
@@ -0,0 +1,99 @@
...
+pub enum QueryMsg {
+    /// Returns the configuration info as a [`QueryConfigResponse`].
+    Config {},
+    /// Returns the event info as a [`QueryEventInfoResponse`].
+    EventInfo {},
+}
+
+/// Response to [`QueryMsg::Config`].
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct QueryConfigResponse {
```
dadamu, line 72: I think we can return `config` instead of separately fields.

contracts/poap/src/msg.rs
```diff
@@ -0,0 +1,99 @@
...
+    /// Id of the cw721 contract that this contract has initialized.
+    pub cw721_contract_code: Uint64,
+    /// Address of the cw721 contract that this contract is using to
+    /// mint the poaps.
+    pub cw721_contract: Addr,
+}
+
+/// Response to [`QueryMsg::EventInfo`].
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct QueryEventInfoResponse {
```
dadamu, line 90: as above we can return `event_info` instead of separately fields

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,443 @@
...
+    let cw721_submessage = SubMsg::reply_on_success(
+        wasm_instantiate(
+            msg.cw721_code_id.into(),
+            &Cw721InstantiateMsg {
+                name: msg.cw721_initiate_msg.name,
+                symbol: msg.cw721_initiate_msg.symbol,
+                minter: env.contract.address.to_string(),
+            },
+            info.funds,
+            "poap cw721".to_string(),
```
dadamu, line 126:
```suggestion
            "poap_cw721".to_string(),
```

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,443 @@
...
+        )?,
+        INSTANTIATE_CW721_REPLY_ID,
+    );
+
+    Ok(Response::new()
+        .add_attribute("method", "instantiate")
```
dadamu, line 132: It would be better to build const value to replace attribute string, say, `"action"` into `ATTRIBUTE_ACTION`, `"creator"` into `ATTRIBUTE_CREATOR`.

contracts/poap/src/contract.rs
bragaz: I agree with the introduction of the `validate()` function for the msg fields checks.

contracts/poap/src/state.rs
bragaz: Yes I think it can be merged without problems @manu0466

contracts/poap/src/contract_tests.rs
```diff
@@ -0,0 +1,962 @@
...
+    fn enable_mint() {
+        let (mut app, poap_contract_addr) = proper_instantiate();
+
+        let msg = ExecuteMsg::EnableMint {};
+        app.execute_contract(Addr::unchecked(ADMIN), poap_contract_addr, &msg, &vec![])
+            .unwrap();
+    }
+
+    #[test]
+    fn normal_user_can_t_enable_mint() {
```
dadamu, line 423:
```suggestion
    fn normal_user_can_not_enable_mint() {
```

contracts/poap/src/contract_tests.rs
```diff
@@ -0,0 +1,962 @@
...
+    fn disable_mint() {
+        let (mut app, poap_contract_addr) = proper_instantiate();
+
+        let msg = ExecuteMsg::DisableMint {};
+        app.execute_contract(Addr::unchecked(ADMIN), poap_contract_addr, &msg, &vec![])
+            .unwrap();
+    }
+
+    #[test]
+    fn normal_user_can_t_disable_mint() {
```
dadamu, line 441:
```suggestion
    fn normal_user_can_not_disable_mint() {
```

contracts/poap/src/contract_tests.rs
```diff
@@ -0,0 +1,962 @@
...
+        );
+        assert!(init_result.is_err());
+    }
+
+    #[test]
+    fn enable_mint() {
+        let (mut app, poap_contract_addr) = proper_instantiate();
+
+        let msg = ExecuteMsg::EnableMint {};
+        app.execute_contract(Addr::unchecked(ADMIN), poap_contract_addr, &msg, &vec![])
```
dadamu, line 418: the state should be checked after executing.

contracts/poap/src/contract_tests.rs
```diff
@@ -0,0 +1,962 @@
...
+        let result = app.execute_contract(Addr::unchecked(USER), poap_contract_addr, &msg, &vec![]);
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn disable_mint() {
+        let (mut app, poap_contract_addr) = proper_instantiate();
+
+        let msg = ExecuteMsg::DisableMint {};
+        app.execute_contract(Addr::unchecked(ADMIN), poap_contract_addr, &msg, &vec![])
```
dadamu, line 436: as above

contracts/poap/src/contract_tests.rs
```diff
@@ -0,0 +1,962 @@
...
+    fn normal_user_can_t_disable_mint() {
+        let (mut app, poap_contract_addr) = proper_instantiate();
+
+        let msg = ExecuteMsg::DisableMint {};
+        let result = app.execute_contract(Addr::unchecked(USER), poap_contract_addr, &msg, &vec![]);
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn mint() {
```
dadamu, line 450: It can split into 4 functions
1. mint_not_started_event_poap_error
2. mint_terminated_event_poap_error
3. mint_without_permission_error
4. mint_with_permission_properly (need a state check)

contracts/poap/src/contract_tests.rs
```diff
@@ -0,0 +1,962 @@
...
+            poap_code_id,
+            Addr::unchecked(ADMIN),
+            &init_msg,
+            &[],
+            "Poap contract",
+            None,
+        );
+        assert!(init_result.is_err());
+    }
+
```
dadamu, line 412: Missing a test for `proper_instantiate` with `config`, `event_info` and etc. state check.

contracts/poap/src/contract_tests.rs
```diff
@@ -0,0 +1,962 @@
...
+            Addr::unchecked(ADMIN),
+            poap_contract_addr.clone(),
+            &msg,
+            &vec![],
+        )
+        .unwrap();
+    }
+
+    #[test]
+    fn only_creator_can_change_event_info() {
```
dadamu, line 676: It can be split into
1. non_creator_change_event_info_error
2. creator_change_event_info_properly ( need a state check on event info )

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,443 @@
...
+) -> Result<Response, ContractError> {
+    let mut event_info = EVENT_INFO.load(deps.storage)?;
+
+    // Check that is the event creator that is changing the event time frame
+    if event_info.creator != info.sender {
+        return Err(ContractError::Unauthorized {});
+    }
+
+    // Check that the event is not ended
+    if event_info.is_ended(&env.block.time) {
```
dadamu, line 300: It can be removed since `end_time` must be after the `start_time`

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,443 @@
...
+    // Check that the start time is not already passed
+    if env.block.time.ge(&start_time) {
+        return Err(ContractError::StartTimeBeforeCurrentTime {
+            current_time: env.block.time,
+            start_time,
+        });
+    }
+
+    // Check that the end time is not already passed
+    if env.block.time.ge(&end_time) {
```
dadamu, line 332: as above, can be removed since it must be after the `start_time`

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,443 @@
...
+        return Err(ContractError::EndTimeBeforeCurrentTime {
+            current_time: env.block.time,
+            end_time,
+        });
+    }
+
+    // Update the event info
+    event_info.start_time = start_time;
+    event_info.end_time = end_time;
+    EVENT_INFO.save(deps.storage, &event_info)?;
```
dadamu, line 342: Can we use `EVENT_INFO.update` instead?

dadamu:
After digging more about `cw_multi_test`. I think it is for `integration_test` between contracts, meaning that it checks the method which involves multiple contracts.
Say, in the poap contract, `instantiate()`, `mint()` and `mint_to()` needs a integration test with cw721. The rest should be tested by unit tests.

contracts/poap/src/msg.rs
```diff
@@ -0,0 +1,99 @@
...
+    },
+    /// Allows the contract's admin to transfer the admin rights to another user.
+    UpdateAdmin { new_admin: String },
+    /// Allows the contract's admin to transfer the minting rights to another user.
+    UpdateMinter { new_minter: String },
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum QueryMsg {
```
dadamu, line 63: It would be useful to have `MintedAmount{ user: String }` to get the minted amount for a user.

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,443 @@
+use crate::error::ContractError;
```
dadamu, line 1: We should have unit tests for contracts. `contract_tests` should be for checking the state outside of this contract.

contracts/poap/src/contract.rs
manu0466: Using update in this case is less efficent then just calling `save`. 
With update you are actualy calling another time `load`, then the lambda passed as argument that performs the modification and finaly `save`.  Since we have already loaded `EventInfo` we can just update the previosly loaded object and save it.

contracts/poap/src/state.rs
manu0466: While performing the changes i noticed that this requires to store the cw721 address inside config as `Option` adding some boilerplate code to check that this field is not empty. So i prefer to keep them separate.

contracts/poap/src/contract_tests.rs
manu0466: `proper_instantiate` is an utility that provides a correclty initialized enviroement. If the initialization fails the test fail providing the reason and a stack trace since is using `unwrap` [here](https://github.com/desmos-labs/desmos-contracts/blob/909561e5c50798c81651c68c07a823b5f8a1a111/contracts/poap/src/contract_tests.rs#L105).
The tests to check the initialization with invalid params are covered in the `instantiate_...` tests.

contracts/poap/src/state.rs
dadamu: @manu0466 Can it be empty address instead of `Option`? I am thinking about that the `instantiate message` should be successful with the proper `instantiate reply`, finally, the cw721 address must not empty after the reply. It can be checked with integration test in my opinion.

manu0466: I agree that the address will be stored since the instantiation is successful only if also the cw721 is properly instantiate but storing the address as empty string just to have an unique `Config` struct seems a dirt hack to me. What are the actual pros of this? We can have a the `QueryConfigResponse` returning just the `Config` struct but beside that i don't see anything else

contracts/poap/src/msg.rs
manu0466: I don't like the idea of returing a struct used to store the contratc state, if in the future we need to extend that struct with some informations that are not required to queried from the client? Also this create an unecessary nesting that a client need to handle while deserializing and process the response

contracts/poap/src/contract_tests.rs
dadamu: Yes, `unwrap` check the initialization fails, but the state check on cw721 is missing. Say, we should check the cw721 `minter` is poap contract  after initialization.

contracts/poap/src/msg.rs
dadamu: Can you give me a sample the information that are not required to client? I think we should return all the config for the user who would get contract state.

Besides, we can reduce some code inside the query from [current one](https://github.com/desmos-labs/desmos-contracts/pull/35/files/03f02fdf42d61d27c22e63731da4d943667feae4#diff-2b9bf8d5b9ea44cca10b66d24526301c233a9e13e8c6dbc3debbb6c0b460b2d9R407-R413) to be like:
```rust
fn query_config(deps: Deps) -> StdResult<QueryConfigResponse> {
    let config = CONFIG.load(deps.storage)?;
    Ok(QueryConfigResponse { config })
}
```

manu0466: The `QueryMsg::EventInfo` don't returns the whole `EventInfo` struct stored inside the state but all the fields except `base_poap_uri`.  
Here the [ADR](https://github.com/desmos-labs/desmos-contracts/blob/master/docs/architecture/adr-001-poap-contract.md#eventinfo-1) reference.

contracts/poap/src/contract_tests.rs
manu0466: Totally forgot that we override the minter address during initialization, nice catch!
I will add a test that after calling `proper_instantiate` checks the cw721 and poap contract state.

contracts/poap/src/state.rs
dadamu: I am just thinking to reduce some code by merging it into config.  Agree with you, we should avoid dirt hack.

contracts/poap/src/msg.rs
dadamu: You are right. Making it separately has more flexibility. Let's remain the current structure.

contracts/poap/src/contract.rs
manu0466: Yes, that control exists only to return the corect error to the client otherwise the contract would return the `EventStarted` error even when the event is finished.

contracts/poap/src/contract.rs
manu0466: As above, also this check should be before the check of `start_time`

contracts/poap/src/msg.rs
```diff
@@ -0,0 +1,170 @@
...
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum ExecuteMsg {
+    /// Allows the contract's admin to enable the [`ExecuteMsg::Mint`].
+    EnableMint {},
+    /// Allows the contract's admin to disable the [`ExecuteMsg::Mint`].
+    DisableMint {},
+    /// If the mint is enabled, allow the user to mint the poap by themself.
+    /// It's disabled after the event's end.
```
bragaz, line 46:
```suggestion
    /// It's disabled before the start of the event and after the event's end.
```

contracts/poap/src/msg.rs
```diff
@@ -0,0 +1,170 @@
...
+    Mint {},
+    /// Allows the contract's admin or the minter to mint a POAP for a specific recipient.
+    /// It's disabled after the event's end.
```
bragaz, line 49:
```suggestion
    /// It's disabled before the start of the event and after the event's end.
```

contracts/poap/src/integration_tests.rs
```diff
@@ -0,0 +1,370 @@
...
+    }
+
+    fn mock_app() -> App {
+        AppBuilder::new()
+            .with_block(BlockInfo {
+                height: 42,
+                time: Timestamp::from_seconds(INITIAL_BLOCK_TIME_SECONDS),
+                chain_id: "testchain".to_string(),
+            })
+            .build(|router, _, storage| {
```
dadamu, line 37: The test does not require any custom bank setting, it can be empty as:
```rust
.build(|_, _, _| {});
```

contracts/poap/src/integration_tests.rs
```diff
@@ -0,0 +1,349 @@
...
+                },
+            )
+            .unwrap();
+
+        assert_eq!(Addr::unchecked(USER), response.user);
+        assert_eq!(1, response.amount);
+    }
+
+    #[test]
+    fn mint_limit() {
```
dadamu, line 187: I think this test should be handled in contract unit test.

contracts/poap/src/integration_tests.rs
```diff
@@ -0,0 +1,349 @@
...
+            Addr::unchecked(MINTER),
+            poap_contract_addr.clone(),
+            &msg,
+            &vec![],
+        );
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn mint_to_limit() {
```
dadamu, line 246: as above.

contracts/poap/src/integration_tests.rs
```diff
@@ -0,0 +1,349 @@
...
+            poap_contract_addr.clone(),
+            &msg,
+            &vec![],
+        );
+        // Should fail since the user have already received the max allowed poaps.
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn mint_to_only_for_minter_and_admin() {
```
dadamu, line 306: as above.

contracts/poap/src/integration_tests.rs
```diff
@@ -0,0 +1,349 @@
...
+            .query_wasm_smart(
+                &poap_contract_addr,
+                &QueryMsg::MintedAmount {
+                    user: USER.to_string(),
+                },
+            )
+            .unwrap();
+
+        assert_eq!(Addr::unchecked(USER), response.user);
+        assert_eq!(1, response.amount);
```
dadamu, line 183: The state of cw721 side should be checked by `Cw721QueryMsg::Tokens{ owner }`.

contracts/poap/src/integration_tests.rs
```diff
@@ -0,0 +1,349 @@
...
+
+        let poap_config: QueryConfigResponse = querier
+            .query_wasm_smart(&poap_contract_addr, &QueryMsg::Config {})
+            .unwrap();
+
+        assert_eq!(Addr::unchecked(ADMIN), poap_config.admin);
+        assert_eq!(Addr::unchecked(MINTER), poap_config.minter);
+        assert_eq!(false, poap_config.mint_enabled);
+        // 1 since is the first uploaded.
+        assert_eq!(Uint64::new(1), poap_config.cw721_contract_code);
```
dadamu, line 118: I think it should be tested in contract test. The integration test should be focus on the states of other contracts.

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,1086 @@
...
+            ContractError::StartTimeBeforeCurrentTime {
+                current_time: env.block.time,
+                start_time: Timestamp::from_seconds(start)
+            },
+            init_result.unwrap_err()
+        );
+    }
+
+    #[test]
+    fn instantiate_with_event_start_equal_current_time() {
```
dadamu, line 514: It is duplicated with `instantiate_with_event_start_before_current_time`. It can be replaced into one test which current block time is equal to start time which is considered as `before current time`.

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,1086 @@
...
+        let env = mock_env();
+        let info = mock_info(ADMIN, &vec![]);
+
+        let result = execute(deps.as_mut(), env, info, msg);
+        // Admin should not be authorized to update the event info
+        assert_eq!(Unauthorized {}, result.unwrap_err());
+    }
+
+    #[test]
+    fn event_info_update_only_if_event_not_started_or_ended() {
```
dadamu, line 781: Can it be splitted into multiple tests to ensure that one test for a check? say:
1. event_info_update_after_event_started_error
2. event_info_update_after_event_terminated_error

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,1086 @@
...
+        );
+
+        // Current time is before event started
+        env.block.time = Timestamp::from_seconds(EVENT_START_SECONDS - 100);
+
+        execute(deps.as_mut(), env, info, msg).unwrap();
+    }
+
+    #[test]
+    fn invalid_event_info() {
```
dadamu, line 849: This test can be handled in `state` or `msg`.

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,1086 @@
...
+
+        let msg = ExecuteMsg::EnableMint {};
+        execute(deps.as_mut(), env, info, msg).unwrap();
+
+        let config = CONFIG.load(&deps.storage).unwrap();
+        assert_eq!(true, config.mint_enabled);
+    }
+
+    #[test]
+    fn normal_user_can_not_enable_mint() {
```
dadamu, line 690:
```suggestion
    fn enable_mint_without_permission_error() {
```

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,1086 @@
...
+
+        let mut init_msg = get_valid_init_msg(1);
+        init_msg.event_info.creator = "a".to_string();
+
+        let result = instantiate(deps.as_mut(), env, info, init_msg);
+        assert!(result.is_err())
+    }
+
+    #[test]
+    fn instantiate_with_event_start_before_current_time() {
```
dadamu, line 492: This test can be moved to `msg`.

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,1086 @@
...
+            ContractError::StartTimeBeforeCurrentTime {
+                current_time: Timestamp::from_nanos(start),
+                start_time: Timestamp::from_nanos(start)
+            },
+            init_result.unwrap_err()
+        );
+    }
+
+    #[test]
+    fn instantiate_with_event_end_before_current_time() {
```
dadamu, line 535: as above

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,1086 @@
...
+            ContractError::EndTimeBeforeCurrentTime {
+                current_time: env.block.time,
+                end_time: Timestamp::from_seconds(end)
+            },
+            init_result.unwrap_err()
+        );
+    }
+
+    #[test]
+    fn instantiate_with_event_end_equal_current_time() {
```
dadamu, line 559: as above

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,1086 @@
...
+            ContractError::StartTimeAfterEndTime {
+                start: Timestamp::from_seconds(start),
+                end: Timestamp::from_seconds(start),
+            },
+            init_result.unwrap_err()
+        );
+    }
+
+    #[test]
+    fn instantiate_with_invalid_poap_uri() {
```
dadamu, line 627: as above

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,1086 @@
...
+        let mut init_msg = get_valid_init_msg(1);
+
+        init_msg.event_info.base_poap_uri = "https://random_domain.com".to_string();
+
+        let init_result = instantiate(deps.as_mut(), env, info, init_msg);
+        assert_eq!(ContractError::InvalidPoapUri {}, init_result.unwrap_err());
+    }
+
+    #[test]
+    fn instantiate_with_invalid_event_uri() {
```
dadamu, line 651: as above

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,1086 @@
...
+            ContractError::StartTimeAfterEndTime {
+                start: Timestamp::from_seconds(start),
+                end: Timestamp::from_seconds(end),
+            },
+            init_result.unwrap_err()
+        );
+    }
+
+    #[test]
+    fn instantiate_with_event_start_eq_end() {
```
dadamu, line 605: as above

contracts/poap/src/integration_tests.rs
manu0466: This tests the cw721 state after initialization and ensure that also the poap contract have a valid state after initizalization, also this contract in order to be properly initialized needs to receive a `MsgInstantiateContractResponse` as a replay data, so imo this is an integration test.

If we want to convert this into a unit test we need to manually call `replay` and serialize `MsgInstantiateContractResponse` but looking into cw-utils I can't find any utility function to properly serialize it, we can obviosly write our own proto serialization but seems just a dirt hack imo, if in the future they change this struct all our tests will probably start failing and we also need to manage some code that do the serialization of a external library struct.

contracts/poap/src/contract.rs
manu0466: This ensure that also that the time equal branch is covered instead of just the before. I do agree that is very similar to the other but is just another test to esure that when we receive a time that is equal to the current one the contract handles it like is a time before the current time.

contracts/poap/src/contract.rs
manu0466: I think that is related to contract not to `msg` or `state` since this ensure that the contract correctly discard invalid `EventInfo`  when calling `UpdateEventInfo`.

contracts/poap/src/contract.rs
manu0466: I can write the tests that ensure that the `validate` method of `InstantiateMsg` and `ExecuteMsg` is implemented correctly but I do not understend how you wanna remove the tests.

contracts/poap/src/integration_tests.rs
dadamu: Yeah, you are right, it is needed a `MsgInstantiateContractResponse`. I covered these reply test in the POAP manager contract with [structure](https://github.com/desmos-labs/desmos-contracts/blob/paul/DCD-51/poap-manager-impl/contracts/poap-manager/src/contract.rs#L238) and reply test[ here](https://github.com/desmos-labs/desmos-contracts/blob/paul/DCD-51/poap-manager-impl/contracts/poap-manager/src/contract.rs#L390). What do you think for this way?

contracts/poap/src/contract.rs
dadamu: As you said, these test can be implemented with `validate` in `msg.rs`, so I think they can be moved to there. As a result, in `contract.rs` only need to have a test called `instantiate_with_invalid_msg_error`.

contracts/poap/src/msg.rs
```diff
@@ -0,0 +1,398 @@
...
+                event_uri: "ipfs://domain.com".to_string(),
+            },
+        };
+
+        assert_eq!(
+            ContractError::StartTimeAfterEndTime { start, end },
+            msg.validate().unwrap_err()
+        );
+
+        // Test start time eq end time
```
dadamu, line 208: I think it can be splitted to another test something like `instantiate_with_invalid_time_start_time_eq_end_time_error`.

contracts/poap/src/msg.rs
```diff
@@ -0,0 +1,398 @@
...
+
+        assert_eq!(
+            ContractError::StartTimeAfterEndTime { start, end },
+            msg.validate().unwrap_err()
+        );
+    }
+
+    #[test]
+    fn instantiate_with_invalid_per_address_limit_error() {
+        // Test start time after end time
```
dadamu, line 238:
```suggestion
```

contracts/poap/src/msg.rs
```diff
@@ -0,0 +1,398 @@
...
+                event_uri: "ipfs://domain.com".to_string(),
+            },
+        };
+
+        assert_eq!(
+            ContractError::InvalidPoapUri {},
+            msg.validate().unwrap_err()
+        );
+
+        // Non ipfs uri
```
dadamu, line 344: I think it can be splitted to `instantiate_with_invalid_base_poap_uri_non_ipfs_error`.

contracts/poap/src/msg.rs
```diff
@@ -0,0 +1,398 @@
...
+            start_time: start.clone(),
+            end_time: end.clone(),
+        };
+
+        assert_eq!(
+            ContractError::StartTimeAfterEndTime { start, end },
+            msg.validate().unwrap_err()
+        );
+
+        // Start time eq end time
```
dadamu, line 385: As above, I think it can be splitted to `update_event_info_start_time_eq_end_time_error`

dadamu:
I think the unit test should follow one test for a one check. In addition, showing that `*_error` if the execution is expected as error, in the other hand, `*_properly` if the execution is expected as success.
Say:
```rust
#[test]
fn mint_with_invalid_msg_error(){
   ...
   assert!(result.is_err())
}

#[test]
fn mint_without_permissions_error(){
   ...
   assert!(result.is_err())
}

#[test]
fn mint_properly(){
   ...
   assert!(result.is_ok())
}
```

What do you think?

contracts/poap/src/contract.rs
dadamu: Okay, I see. Can we split it into different tests?
1. invalid msg
2. start time before current time
3. start time eq current time
4. end time before current time
5. end time eq current time

contracts/poap/src/integration_tests.rs
manu0466: I see, as I said this aproach can lead to problems in the future if the `MsgInstantiateContractResponse` changes.
We have the integration tests framework that takes care of this for apart for a strong motivation i don't see why don't use that.

bragaz: I think it's okay to test the reply working properly (if not already tested by `cosmwasm` itself) but I think it's useful to have also this integration test since we are calling different components and not testing units only. I would leave this test here if possible.

manu0466:
> I think the unit test should follow one test for a one check. In addition, showing that `*_error` if the execution is expected as error, in the other hand, `*_properly` if the execution is expected as success. Say:
> 
> ```rust
> #[test]
> fn mint_with_invalid_msg_error(){
>    ...
>    assert!(result.is_err())
> }
> 
> #[test]
> fn mint_without_permissions_error(){
>    ...
>    assert!(result.is_err())
> }
> 
> #[test]
> fn mint_properly(){
>    ...
>    assert!(result.is_ok())
> }
> ```
> 
> What do you think?

I agree on the naming convention for the tests, i suggest to use `unwrap` instead of `assert!` since provides more details when the result is an error.  
For the error side I suggest, where is possible, to use `assert_eq!` with the actual error to ensure that on failure we have the expected error.

contracts/poap/src/integration_tests.rs
dadamu: I see, let's remain the reply msg check in integration test.

contracts/poap/src/msg.rs
```diff
@@ -0,0 +1,401 @@
...
+    /// Address for which the request was made.
+    pub user: Addr,
+    /// Amount of poaps minted from the user.
+    pub amount: u32,
+}
+
+impl InstantiateMsg {
+    /// Checks that the data inside the message are coherent.
+    /// NOTE: This function don't checks if the address are valid.
+    pub fn validate(&self) -> Result<(), ContractError> {
```
dadamu, line 117: cw721 code id should be checked not to be 0 here as well.

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,1375 @@
...
+        let mut deps = mock_dependencies();
+        let env = mock_env();
+        let info = mock_info(ADMIN, &vec![]);
+        let mut init_msg = get_valid_init_msg(1);
+
+        // Invalid uri
+        init_msg.event_info.base_poap_uri = "invalid_uri".to_string();
+
+        let init_result = instantiate(deps.as_mut(), env, info, init_msg);
+        assert!(init_result.is_err());
```
dadamu, line 643:
```suggestion
assert_eq!(ContractError::InvalidPoapUri {}, init_result.unwrap_err());
```

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,1375 @@
...
+        let mut deps = mock_dependencies();
+        let env = mock_env();
+        let info = mock_info(ADMIN, &vec![]);
+        let mut init_msg = get_valid_init_msg(1);
+
+        // Invalid uri
+        init_msg.event_info.event_uri = "invalid_uri".to_string();
+
+        let init_result = instantiate(deps.as_mut(), env, info, init_msg);
+        assert!(init_result.is_err());
```
dadamu, line 670:
```suggestion
assert_eq!(ContractError::InvalidEventUri {}, init_result.unwrap_err());
```

contracts/poap/src/msg.rs
manu0466: I think is not needed to add an extra error code for that, if an user provides an invalid contract code ther will be the wasm instantiate message that will fail. What you think?

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,1375 @@
...
+            ContractError::EndTimeBeforeCurrentTime {
+                current_time: Timestamp::from_nanos(end),
+                end_time: Timestamp::from_nanos(end),
+            },
+            init_result.unwrap_err()
+        );
+    }
+
+    #[test]
+    fn instantiate_with_event_start_after_end_error() {
```
dadamu, line 587: These test cases can be merged to be `instantiate_with_invalid_msg_error` with one case since they are well tested inside `msg.validate`:
1. instantiate_with_event_start_after_end_error
2. instantiate_with_event_start_equal_end_error
3. instantiate_with_invalid_poap_uri_error
4. instantiate_with_non_ipfs_poap_uri_error
5. instantiate_with_invalid_event_uri_error
6. instantiate_with_non_ipfs_event_uri_error

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,1375 @@
...
+        };
+
+        // Current time is before event started
+        env.block.time = Timestamp::from_seconds(EVENT_START_SECONDS - 100);
+
+        execute(deps.as_mut(), env, info, msg).unwrap();
+    }
+
+    #[test]
+    fn event_info_start_time_equal_end_time_error() {
```
dadamu, line 899: Similar as above, `event_info_start_time_equal_end_time_error`, `event_info_start_time_after_end_time_error` can be merged into `update_event_info_with_invalid_msg_error` with one case.

contracts/poap/src/contract.rs
manu0466: This test was a single case before and then got splitted 😅
For consistence I will prefere to keep a single test for each case

contracts/poap/src/contract.rs
manu0466: As above also these tests were also a single one and then they were split

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,1375 @@
...
+            ContractError::StartTimeAfterEndTime {
+                start: Timestamp::from_seconds(INITIAL_BLOCK_TIME_SECONDS + 2),
+                end: Timestamp::from_seconds(INITIAL_BLOCK_TIME_SECONDS),
+            },
+            result.unwrap_err()
+        );
+    }
+
+    #[test]
+    fn update_admin_only_from_admin() {
```
dadamu, line 1044: I think it can be splited to:
1. update_admin_without_permission_error
2. update_admin_with_permission_properly

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,1375 @@
...
+        let info = mock_info(ADMIN, &vec![]);
+
+        execute(deps.as_mut(), env.clone(), info, msg.clone()).unwrap();
+
+        let config = CONFIG.load(&deps.storage).unwrap();
+        assert_eq!(NEW_ADMIN, config.admin.as_str());
+    }
+
+    #[test]
+    fn update_minter_only_from_admin() {
```
dadamu: as above, it can be splited into:
1. update_minter_without_permission_error
2. update_minter_with_permission_properly

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,1375 @@
...
+        assert_eq!(
+            ContractError::MaxPerAddressLimitExceeded {
+                recipient_addr: USER.to_string()
+            },
+            response.unwrap_err()
+        );
+    }
+
+    #[test]
+    fn mint_to_from_user_error() {
```
dadamu, line 1345:
```suggestion
    fn mint_to_without_permission_error() {
```

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,1375 @@
...
+
+        let msg = ExecuteMsg::Mint {};
+        let result = execute(deps.as_mut(), env.clone(), info, msg);
+
+        // Event is not started
+        assert_eq!(ContractError::MintDisabled {}, result.unwrap_err());
+    }
+
+    #[test]
+    fn mint_limit() {
```
dadamu, line 1215: It can be splitted into:
1. mint_token_properly
2. mint_out_of_max_amount_error

contracts/poap/src/contract.rs
```diff
@@ -0,0 +1,1375 @@
...
+        assert_eq!(
+            ContractError::MaxPerAddressLimitExceeded {
+                recipient_addr: USER.to_string()
+            },
+            response.unwrap_err()
+        );
+    }
+
+    #[test]
+    fn mint_to_limit() {
```
dadamu, line 1278: It could be removed since it is duplicated with `mint_limit`

contracts/poap/src/contract.rs
dadamu: Okay, just thinking about they are covered by `msg.validate` already so we can reduce some tests in the `contract.rs`. I am fine with remaining them.

contracts/poap/src/contract.rs
manu0466: The `mint_token_properly` is already coverd inside the integration tests.
This can just be renamed into `mint_out_of_max_amount_error`

contracts/poap/src/contract.rs
manu0466: If we remove this test we don't ensure that the `MintTo` message don't allow to mint more then `per_address_limit` poap.
This test can be renamed into `mint_to_out_of_max_amount_error` and add the missing integration test `mint_to_with_permission_properly`.

dadamu: Yes, my bad, I forgot that `mint_to` and `mint` are different messages.

dadamu (approved):
Great job! Looks good to me.

contracts/poap/src/contract.rs
manu0466: I understand, but a test imo should ensure that a certain functionality behaves correctly regardless of how it is written internally. If we have just one test we are making an implicit assumption that the `validate` function will be called

dadamu: Yeah, you are right. Calling `msg.validate` is not the thing must to do in the contract, unlike chain app must run `msg.ValidateBasic` before executing it. I am mixing up them.

bragaz (approved):
Great Job, Let's merge this! 🚀

PR #42, opened 2022-08-11
ci: splitted lint and test jobs in 2 workflows

manu0466:
* Moved the lint tests into a dedicated workflow file
* Moved the contracts tests into a dedicated workflow file
* Reduced the number of containers spawned to decrease the execution time


dadamu (approved):
Nice improvement!

bragaz (approved):

PR #43, opened 2022-08-25
docs(adr): ADR-004 Tips contract

bragaz:
This PR contains the ADR describing the implementation specs for a Tips contract.

### DIFF
--- a/docs/architecture/adr-004-tips-contract.md
+++ b/docs/architecture/adr-004-tips-contract.md
@@ -1,0 +1,72 @@
+# ADR 003: Remarkables Contract
+
+## Changelog
+
+- Aug 24, 2022: Initial draft.
+
+## Status
+DRAFTED
+
+## Abstract
+This ADR defines the architecture of the Tips contract. This contract manage sending and tracking of tips made within
+Desmos based applications and exposes queries that allow to gather all users sent and received tips easily.
+
+## Context
+Desmos based applications may want to leverage on the web3 features and one of the most popular one is for sure the possibility to tip users. Currently, it's easy to integrate a feature that allow an application to send tokens as tips between users but very difficult to track those tips in an easy way.
+Plus application creators might want to integrate some form of monetization that allows them to earn a little percentage when user send tips to each other.
+
+## Decision
+The idea here is to write a contract called `Tips`, that handles the sending and tracking of user's tips.
+This contract allows sending tips between users as well tracking sent and received tips for each one of them.
+The tips tracking should be limited to a maximum tracked record number that helps to keep the contract storage as small
+as possible. Also, it doesn't make much sense to store all the tips record that a user received because each one of them will
+most likely check only the latest received.
+Additionally, the contract allows its admin to set a fee that users has to pay in order to be allowed to send a tip. This fee, if not set, is equal to zero.
+
+## Specifications
+Here below the specifications for the contract's messages:
+
+### Messages
+
+#### Instantiate
+```rust
+pub struct InstantiateMsg {
+  pub admin: String,
+  pub subspace_id: u64,
+  pub service_fee: Vec<Coin>,
+}
+```
+
+* The `admin` identifies the user that controls the contract;
+* The `subspace_id` identifies the application which is deploying the contract;
+* The `service_fee` identifies a fee that the users need to pay to use the contract.
+
+#### Execute
+```rust
+pub enum ExecuteMsg{
+  SendTip{},
+  UpdateServiceFee{},
+  UpdateAdmin{},
+}
+```
+
+### Query
+```rust
+pub enum QueryMsg {
+  /// Return a ConfigResponse containing the configuration info of the contract
+  Config{},
+  ReceivedTips{},
+  SentTips{}
+}
+```
+
+#### Config
+The `Config{}` query returns the contract's configuration inside a `ConfigResponse`.
+```rust
+pub struct QueryConfigResponse {
+  pub admin: Addr,
+  pub subspace_id: u64,
+  pub service_fee: Vec<Coin>,
+}
+```
+
--- a/docs/architecture/adr-004-tips-contract.md
+++ b/docs/architecture/adr-004-tips-contract.md
@@ -1,8 +1,8 @@
-# ADR 003: Remarkables Contract
+# ADR 004: Tips Contract
 
 ## Changelog
 
-- Aug 24, 2022: Initial draft.
+- Aug 25, 2022: Initial draft.
 
 ## Status
 DRAFTED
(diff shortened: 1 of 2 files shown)


docs/architecture/adr-004-tips-contract.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Changelog
+
+- Aug 25, 2022: Initial draft.
+
+## Status
+DRAFTED
+
+## Abstract
+This ADR defines the architecture of the Tips contract. This contract manage sending and tracking of tips made within
```
RiccardoM, line 11:
```suggestion
This ADR defines the architecture of the Tips contract. This contract manages sending and tracking of tips made within
```

docs/architecture/adr-004-tips-contract.md
```diff
@@ -0,0 +1,141 @@
...
+Desmos based applications and exposes queries that allow to gather all users sent and received tips easily.
+
+## Context
+Desmos based applications may want to leverage on the web3 features and one of the most popular one is for sure the possibility to tip users. Currently, it's easy to integrate a feature that allow an application to send tokens as tips between users but very difficult to track those tips in an easy way.
```
RiccardoM, line 15:
```suggestion
Desmos based applications may want to leverage on the web3 features and one of the most popular one is for sure the possibility to tip users. Currently, it's easy to integrate a feature that allows sending tokens as tips between users, but it's very difficult to track those tips in an easy way.
```

docs/architecture/adr-004-tips-contract.md
```diff
@@ -0,0 +1,141 @@
...
+Plus application creators might want to integrate some form of monetization that allows them to earn a little percentage when user send tips to each other.
```
RiccardoM, line 16:
```suggestion
Additionally, application developers might want to integrate some form of monetization that allows them to earn a little percentage when users send tips to each other.
```

docs/architecture/adr-004-tips-contract.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Decision
+The idea here is to write a contract called `Tips`, that handles the sending and tracking of user's tips.
+This contract allows sending tips between users as well tracking sent and received tips for each one of them.
+The tips tracking should be limited to a maximum tracked record number that helps to keep the contract storage as small
+as possible. Also, it doesn't make much sense to store all the tips record that a user received because each one of them will
+most likely check only the latest received.
+Additionally, the contract allows its admin to set a fee that users has to pay in order to be allowed to send a tip. This fee, if not set, is equal to zero.
```
RiccardoM, line 24:
```suggestion
Additionally, the contract allows its admin to set a fee that users have to pay in order to be allowed to send a tip. This fee, if not set, defaults to zero.
```

docs/architecture/adr-004-tips-contract.md
```diff
@@ -0,0 +1,141 @@
...
+Here below the specifications for the contract's messages:
+
+### Messages
+
+#### Instantiate
+```rust
+pub struct InstantiateMsg {
+  pub admin: String,
+  pub subspace_id: u64,
+  pub service_fee: Vec<Coin>,
```
RiccardoM, line 36: Currently the service fees are set as a list of `Coin`. However, there might be cases in which a percentage fee would be better. Should we allow setting both kind of fees (percentage, and fixed)?

docs/architecture/adr-004-tips-contract.md
```diff
@@ -0,0 +1,141 @@
...
+pub enum ExecuteMsg{
+  SendTip{post_id: u64, receiver: String},
+  UpdateServiceFee{new_fee: Vec<Coin>},
+  UpdateAdmin{new_admin: String},
+  UpdateSavedTipsRecordThreshold{new_threshold: u64}
+}
+```
+
+##### SendTip
+With the `SendTip` message the user call the contract to send a tip to another user to show their support towards a specific content they made.
```
RiccardoM, line 57:
```suggestion
With the `SendTip` message the user can call the contract to send a tip to another user to show their support towards a specific content they made.
```

docs/architecture/adr-004-tips-contract.md
```diff
@@ -0,0 +1,141 @@
...
+The `MessagInfo` field of the message contains both:
+* the tip identified by the `funds` field
+* the sender identified by the `sender` field
+
+Ideally the message should perform the following checks:
+* Check that the post associated with the given `post_id` exists
+* Check that the `receiver` is equal to the post author
+* If there's any service fee, check that it's covered by the tip amount
```
RiccardoM, line 65: If you want to uniquely identify posts you should also specify the subspace id, as there can be multiple posts with the same ids inside different subspaces. 

That being said, why is tipping here only tied to posts? Why not allowing a generic (non-content related) tip?

docs/architecture/adr-004-tips-contract.md
```diff
@@ -0,0 +1,141 @@
...
+
+If the checks pass successfully, then the tip record can be saved. The number of records saved should never surpass the threshold set.
+The record can be saved using the `Map` structure of the `cw_storage_plus` package that allows to:
+* save elements with combined keys (similar to what we do in the storages of Desmos Core modules)
+* iterate through them easily with pre-built iterators
+* faster data fetches
```
RiccardoM, line 71:
```suggestion
* fetch data more efficiently
```

docs/architecture/adr-004-tips-contract.md
```diff
@@ -0,0 +1,141 @@
...
+With this prefix, using the possibilities offer by the `Map`'s iterators it's possible to retrieve:
+* All the post's received tips
+* All the tips sent by a user
+* All the tips received by a user
+
+##### UpdateServiceFee
+With the `UpdateServiceFee` message the user can update the previously set service fee.
+The message should make sure that the user trying to perform the change is either:
+* The contract admin
+* The subspace owner or admin
```
RiccardoM, line 83:
```suggestion
* the contract admin
* the subspace owner or admin
```

Not sure about allowing the subspace owner/admin to change this to be honest. It might be that one DApp uses an already existing subspace to store its data. This might be the case with open subspaces. We shouldn't allow the subspace owner to change the contract specifications in this case

docs/architecture/adr-004-tips-contract.md
```diff
@@ -0,0 +1,141 @@
...
+
+##### UpdateAdmin
+With the `UpdateAdmin` message the user can update the contract's admin.
+The message should make sure that the user trying to make the edit is the actual contract admin.
+
+##### UpdateSavedTipsRecordThreshold
+With the `UpdateSavedTipsRecordThreshold` message the user can update the Tips records threshold.
```
RiccardoM, line 90:
```suggestion
With the `UpdateSavedTipsRecordThreshold` message the user can update the tips records threshold.
```

docs/architecture/adr-004-tips-contract.md
bragaz: Percentage means a % based on the tips the user give? 
If so, I think it can be done without problems

docs/architecture/adr-004-tips-contract.md
bragaz: > If you want to uniquely identify posts you should also specify the subspace id, as there can be multiple posts with the same ids inside different subspaces.

@RiccardoM If you take a look at the `InstantiateMsg{}` it contains also a `subspace_id` field that identifies which subspace the instantiated contract is tied with. This way anyone creating their own social app can decide whether or not to have the tipping feature. So ideally each social can have its own tipping contract (which is identified by an address) that manages tips only for that. `Why` this? 
1. Because relying on only 1 smart contract for all the social networks make the tips managing more complicated and will most likely slow down the execution times of it. 
2. It requires everyone to force  an update of their own client if the contract migrates to a new one because of an update

This is the way with which NFTs contract are deployed. Each collection has its own contract.

> That being said, why is tipping here only tied to posts? Why not allowing a generic (non-content related) tip?

There's no particular reason, I've did it like that because I was looking at how the UI of our product is built and since it doesn't allow any generic tipping I adapted like it. But we can make the `post_id` field optional so we can allow generic tipping with the same message.

docs/architecture/adr-004-tips-contract.md
bragaz: @RiccardoM right, I didn't think about this possibility. So we can make it updatable only from the contract admin

docs/architecture/adr-004-tips-contract.md
RiccardoM: @bragaz Yes. There might be cases in which contract owners want to get a cut on the tips in percentage rather thank fixed value. I think allowing them to decide what kind of fee type to use (percentage or fixed) would be better. In this way we can support multiple use cases.

_Fixed_  
The contract owner sets a fixed amount of service fees to be paid (e.g `0.01` DSM) each time a user wants to send a tip.

_Percentage_ 
The contract owner sets a percentage amount of service fees to be paid (e.g. `10%`) each time a user wants to send a tip. The service fee amount will then be computed as `service_fee_amount = service_fee_percentage * tip_amoun`

docs/architecture/adr-004-tips-contract.md
RiccardoM: > If you take a look at the InstantiateMsg{} it contains also a subspace_id field that identifies which subspace the instantiated contract is tied with

All right, missed that. 

> But we can make the post_id field optional so we can allow generic tipping with the same message.

I think it would make more sense to have something more generic, like a `target` field that can be either `PostTarget` or `UserTarget` or (in the future) any other kind of target. This way we can easily distinguish what was the meaning of a tip

docs/architecture/adr-004-tips-contract.md
```diff
@@ -0,0 +1,142 @@
...
+With this prefix, using the possibilities offer by the `Map`'s iterators it's possible to retrieve:
+* All the post's received tips
+* All the tips sent by a user
+* All the tips received by a user
+
+##### UpdateServiceFee
+With the `UpdateServiceFee` message the user can update the previously set service fee.
+The message should make sure that the user trying to perform the change is either:
+* The contract admin
+* The contract admin
```
RiccardoM, line 84:
```suggestion
The message should make sure that the user tyring to make the edit is the actual contract admin.
```

docs/architecture/adr-004-tips-contract.md
```diff
@@ -0,0 +1,142 @@
...
+The record can be saved using the `Map` structure of the `cw_storage_plus` package that allows to:
+* save elements with combined keys (similar to what we do in the storages of Desmos Core modules)
+* iterate through them easily with pre-built iterators
+* fetch data more efficiently
+
+Ideally the map should have the following key `post_id + sender + receiver` pointing at the tip amount.
+With this prefix, using the possibilities offer by the `Map`'s iterators it's possible to retrieve:
+* All the post's received tips
+* All the tips sent by a user
+* All the tips received by a user
```
RiccardoM, line 78:
```suggestion
* all the post's received tips;
* all the tips sent by a user;
* all the tips received by a user.
```

docs/architecture/adr-004-tips-contract.md
```diff
@@ -0,0 +1,142 @@
...
+Ideally the message should perform the following checks:
+* Check that the post associated with the given `post_id` exists
+* Check that the `receiver` is equal to the post author
+* If there's any service fee, check that it's covered by the tip amount
+
+If the checks pass successfully, then the tip record can be saved. The number of records saved should never surpass the threshold set.
+The record can be saved using the `Map` structure of the `cw_storage_plus` package that allows to:
+* save elements with combined keys (similar to what we do in the storages of Desmos Core modules)
+* iterate through them easily with pre-built iterators
+* fetch data more efficiently
```
RiccardoM, line 72:
```suggestion
* save elements with combined keys (similar to what we do in the storages of Desmos Core modules);
* iterate through them easily with pre-built iterators;
* fetch data more efficiently.
```

RiccardoM (approved):

PR #46, opened 2022-08-30
docs(adr): edit adr-001 poap contract to store claimers

dadamu:
This PR reviews the ADR001 to store the claimer info into cw721-base extension field.

### DIFF
--- a/docs/architecture/adr-001-poap-contract.md
+++ b/docs/architecture/adr-001-poap-contract.md
@@ -147,11 +147,11 @@ The metadata should be filled as the following ones:
 ```
 * The `cw721_code_id` identifies the code of the `CW721-base` contract initialised by this contract.
 
-#### TokenExtInfo
-The `TokenInfo` is used to store the extended information inside CW721-base.
+#### Metadata
+The `Metadata` is used to store the extended information inside CW721-base.
 
 ```rust
-pub struct TokenExtInfo{
+pub struct Metadata {
   pub claimer: Addr,
 }
 ```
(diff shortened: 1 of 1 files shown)


docs/architecture/adr-001-poap-contract.md
```diff
@@ -29,8 +29,8 @@ We will create a `POAP` contract that allows a selected `minter` (ideally anothe
 collection associated with a specific event. The contract will implement the `CW721-base`, the basic implementation of
 NFTs in CosmWasm.
 The `CW721-base` will take care of store all the information related to:
-- Associations between POAPs and attendees;
-- Associations between POAPs and metadata.
```
dadamu, line 33: Metadata does not store in cw721-base, it handled by `EventInfo` in POAP contract now.

manu0466 (approved):

docs/architecture/adr-001-poap-contract.md
dadamu: Nevermind, it still store metadata in cw721-base

docs/architecture/adr-001-poap-contract.md
```diff
@@ -146,6 +147,16 @@ The metadata should be filled as the following ones:
 ```
 * The `cw721_code_id` identifies the code of the `CW721-base` contract initialised by this contract.
 
+#### TokenExtInfo
```
bragaz, line 150: I think we can name this simply as Claimer

docs/architecture/adr-001-poap-contract.md
```diff
@@ -146,6 +147,16 @@ The metadata should be filled as the following ones:
...
+The `TokenInfo` is used to store the extended information inside CW721-base.
+
+```rust
+pub struct TokenExtInfo{
+  pub claimer: Addr,
```
bragaz, line 155:
```suggestion
  pub address: Addr,
```

dadamu: I would have a json like:
```json
{
  "claimer": "desmosabcdefg..."
}
```

docs/architecture/adr-001-poap-contract.md
dadamu: Maybe we should name it into `Extension`, since the field name is `extension` by querying, like the sample of [cw721-metadata](https://github.com/CosmWasm/cw-nfts/tree/main/contracts/cw721-metadata-onchain).
What do you think?

bragaz: What if we name it `Metadata` and we put inside the object the `claimer`?
So that if we need to extend the object with further data it will be easier since we already have the on-chain metadata. What do you think?

dadamu: @bragaz Great, `Metadata` would be more proper.

bragaz (approved):

PR #48, opened 2022-08-31
refactor: fix generated poap-manager json schema

manu0466:
This PR fixes a type name collission in the generated json schema of `poap-manager`.
The `InstantiateMsg` of `poap-manager` contains the `poap` `InstantiateMsg` that contains the `cw721` `InstantiateMsg`  this causes a type name collision between the poap and cw721 `InstantiateMsg` because both have the same type name resulting in a schema that have the wrong type for the field `cw721_initiate_msg` of the poap  `InstantiateMsg` .
To avoid the name collission I used the schemars attribute `rename` to change the type name and the schemars `title` attribute to preserve the file name of the generated json schemas.
[Here](https://github.com/GREsau/schemars/blob/00e482c3a19e7b3e15f8a0de99c7d717bf96595f/docs/1.1-attributes.md#title-description) the documentation related to the schemars attributes.

### DIFF
--- a/contracts/poap-manager/schema/instantiate_msg.json
+++ b/contracts/poap-manager/schema/instantiate_msg.json
@@ -24,7 +24,7 @@
       "description": "Initialization message of the POAP contract.",
       "allOf": [
         {
-          "$ref": "#/definitions/InstantiateMsg"
+          "$ref": "#/definitions/PoapInstantiateMsg"
         }
       ]
     }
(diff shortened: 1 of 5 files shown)


dadamu (commented):
Great to me! I think `EventInfo` also in POAP has naming collision. Can you also fix them?

manu0466:
> Great to me! I think `EventInfo` in POAP has also naming collision. Can you also fix them?
@dadamu I don't understund where the collision happen can you please elaborate a bit more? 
Btw using the schema to generate some ts code i don't have any other collision or wrong type for the instantiate,execute, and query messages

dadamu (approved):

PR #49, opened 2022-09-01
feat: build cw721 poap contract

dadamu:
Since the issue of the original cw721-base has `Empty` as `Extension` instead of our custom `Metadata` with claimer. This PR intends to fix the issue by building our custom cw721 contract for POAP.


contracts/poap-manager/src/test_utils.rs
```diff
@@ -6,6 +6,7 @@ use cw721_base::{
     ContractError as Cw721ContractError, Cw721Contract, ExecuteMsg as Cw721ExecuteMsg,
     InstantiateMsg as Cw721InstantiateMsg, QueryMsg as Cw721QueryMsg,
 };
+use cw721_poap::{Metadata};
```
manu0466, line 9:
```suggestion
use cw721_poap::Metadata;
```
Lint error

contracts/cw721-poap/rustfmt.toml
```diff
@@ -0,0 +1,15 @@
+# stable
```
manu0466, line 1: Can we move this file to the project root directory so that we have a common style for each contract?

dadamu: Removed all rustfmt in the contracts.

manu0466 (approved):
Nice work!

bragaz (approved):
Looks good to me 👍

PR #51, opened 2022-09-02
docs: edit poap adr to expose queries from cw721 contract

dadamu:
This PR edits the adr to expose nft related queries from cw721. It would be easily to integrate from the app.


manu0466 (approved):

dadamu:
Used wrong dcd tag. I will create another PR.

PR #52, opened 2022-09-02
feat: implement edited adr 001 exposing cw721 queries

dadamu:
This PR exposes the nft info queries from cw721 for POAP contract.
deps: #51


dadamu:
Used wrong dcd tag, will create another PR.

PR #57, opened 2022-09-05
docs: updated ADR-004

bragaz:
The ADR has been updated with the following suggestion from @manu0466:
* Added a `decimals` field to help understand the decimal places of the `value` field;
* Updated the `UpdateServiceFee` message to make it able to update a `ServiceFee` object;
* Made the `service_tips_record_threshold` a `u32` to help save disk space from the contract;
* Removed the redundant `receiver` field from the `Target::ContentTarget` to avoid a useless equality check with the post's author;
* Added a `ClaimFees{ recipient: String }` message to allow contract's admin to be able to claim contract's collected fees.

All credits goes to @manu0466.

### DIFF
--- a/docs/architecture/adr-004-tips-contract.md
+++ b/docs/architecture/adr-004-tips-contract.md
@@ -61,7 +61,7 @@ pub enum ServiceFee {
 ```
 
 The `Percentage` fee is represented by a:
-  * `value` field that identifies the number of percentage the admin applies (e.g. value = 2 means that the percentage can be 2%, 20%, etc..based on the 2nd field);
+  * `value` field that identifies the number of percentage the admin applies (e.g. value = 2 means that the percentage can be 2%, 0,2%, etc..based on the 2nd field);
   * `decimals` field represents the number of decimal places that come before the `value`.
 
 The `Percentage` fee will then be calculated as follows so: `tip_amount * value/10^decimals`
--- a/docs/architecture/adr-004-tips-contract.md
+++ b/docs/architecture/adr-004-tips-contract.md
@@ -5,7 +5,8 @@
 - Aug 25, 2022: Initial draft;
 - Aug 30, 2022: First review;
 - Sept 1, 2022: Second review;
-- Sept 5, 2022: Third review.
+- Sept 5, 2022: Third review;
+- Sept 5, 2022: Fourth review.
 
 ## Status
 DRAFTED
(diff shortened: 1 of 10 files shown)


docs/architecture/adr-004-tips-contract.md
```diff
@@ -54,26 +55,31 @@ tip and paid to the contract.
...
   Fixed {amount: Vec<Coin>},
-  Percentage{value: u64}
+  Percentage{value: u64, decimals: u32}
 }
 ```
 
-* The value of the percentage is represented by a `u64` cause float can't be used inside contracts. The Service fee will then be calculated as follow so:
-  * Service Fee = `tip_amount * value/100`
+The `Percentage` fee is represented by a:
+  * `value` field that identifies the number of percentage the admin applies (e.g. value = 2 means that the percentage can be 2%, 20%, etc..based on the 2nd field);
```
manu0466, line 63:
```suggestion
  * `value` field that identifies the number of percentage the admin applies (e.g. value = 2 means that the percentage can be 2%, 0,2%, etc..based on the 2nd field);
```
I think that having 0,2% as second example makes more clear the meaning of `decimals`

manu0466 (approved):

docs/architecture/adr-004-tips-contract.md
```diff
@@ -54,26 +55,31 @@ tip and paid to the contract.
...
 }
 ```
 
-* The value of the percentage is represented by a `u64` cause float can't be used inside contracts. The Service fee will then be calculated as follow so:
-  * Service Fee = `tip_amount * value/100`
+The `Percentage` fee is represented by a:
+  * `value` field that identifies the number of percentage the admin applies (e.g. value = 2 means that the percentage can be 2%, 0,2%, etc..based on the 2nd field);
+  * `decimals` field represents the number of decimal places that come before the `value`.
+
+The `Percentage` fee will then be calculated as follows so: `tip_amount * (value*10^decimals)/(100*10^decimals)`
```
dadamu, line 66: Shouldn't it be `tip_amount * value /  10^decimals`?
value = 2, decimals = 1, then it is 0.2
value = 2, decimals = 2, then it is 0.02

bragaz: @dadamu You are right. Thanks for the correction.

PR #58, opened 2022-09-07
feat: ADR-004 tips contract

manu0466:
This PR implements the `tips` contract as described in the [ADR-004](https://github.com/desmos-labs/desmos-contracts/blob/master/docs/architecture/adr-004-tips-contract.md)

### DIFF
--- a/contracts/tips/src/contract.rs
+++ b/contracts/tips/src/contract.rs
@@ -114,7 +114,7 @@ fn execute_send_tip(
     tip_amount: Vec<Coin>,
 ) -> Result<Response<DesmosMsg>, ContractError> {
     if info.funds.is_empty() {
-        return Err(ContractError::EmptyFounds {});
+        return Err(ContractError::EmptyFunds {});
     }
 
     let config = CONFIG.load(deps.storage)?;
@@ -500,10 +500,10 @@ mod tests {
         deps: DepsMut<DesmosQuery>,
         from: &str,
         to: &str,
-        founds: &[Coin],
+        funds: &[Coin],
         coins: &[Coin],
     ) -> Result<Response<DesmosMsg>, ContractError> {
-        let info = mock_info(from, founds);
+        let info = mock_info(from, funds);
         execute(
             deps,
             mock_env(),
@@ -521,10 +521,10 @@ mod tests {
         deps: DepsMut<DesmosQuery>,
         from: &str,
         post_id: u64,
-        founds: &[Coin],
+        funds: &[Coin],
         coins: &[Coin],
     ) -> Result<Response<DesmosMsg>, ContractError> {
-        let info = mock_info(from, founds);
+        let info = mock_info(from, funds);
         execute(
             deps,
             mock_env(),
@@ -708,7 +708,7 @@ mod tests {
     }
 
     #[test]
-    fn tip_with_empty_founds() {
+    fn tip_with_empty_funds() {
         let mut deps = mock_dependencies_with_custom_querier(&[]);
 
         init_contract(deps.as_mut(), 1, None, 5).unwrap();
@@ -722,7 +722,7 @@ mod tests {
         )
         .unwrap_err();
 
-        assert_eq!(ContractError::EmptyFounds {}, tip_error);
+        assert_eq!(ContractError::EmptyFunds {}, tip_error);
     }
 
     #[test]
--- a/contracts/tips/src/state.rs
+++ b/contracts/tips/src/state.rs
@@ -91,10 +91,10 @@ pub fn tips_record<'a>() -> IndexedMap<'a, String, StateTip, StateTipIndex<'a>> 
 impl StateServiceFee {
     /// Computes the fees that the contract will holds and the coins that
     /// can be sent to the user.
-    /// * `founds` - Coins sent from the user to the contract.
+    /// * `funds` - Coins sent from the user to the contract.
     /// * `tip_amount` - Coins from which to calculate the fees.
-    pub fn check_fees(&self, founds: &[Coin], tip_amount: &[Coin]) -> Result<(), ContractError> {
-        let founds = sum_coins_sorted(founds.to_vec())?;
+    pub fn check_fees(&self, funds: &[Coin], tip_amount: &[Coin]) -> Result<(), ContractError> {
+        let funds = sum_coins_sorted(funds.to_vec())?;
         // Compute the fees
         let mut fee = match self {
             StateServiceFee::Fixed { amount } => amount.clone(),
@@ -112,24 +112,24 @@ impl StateServiceFee {
 
         // Put the tip amount inside the fees
         fee.extend(tip_amount.to_vec());
-        // Check fees + tips < founds
+        // Check fees + tips < funds
         for fee_plus_tip in sum_coins_sorted(fee)?.drain(0..) {
-            // Search the fee coin inside the founds sent to the contract
-            let found_coin_amount = founds
+            // Search the fee coin inside the funds sent to the contract
+            let fund_coin_amount = funds
                 .binary_search_by(|coin| coin.denom.cmp(&fee_plus_tip.denom))
-                .map(|index| founds[index].amount)
+                .map(|index| funds[index].amount)
                 .map_err(|_| ContractError::InsufficientAmount {
                     denom: fee_plus_tip.denom.clone(),
                     requested: fee_plus_tip.amount,
                     provided: Uint128::zero(),
                 })?;
 
-            // Ensure tip amount + fee <= provided founds
-            if fee_plus_tip.amount > found_coin_amount {
+            // Ensure tip amount + fee <= provided funds
+            if fee_plus_tip.amount > fund_coin_amount {
                 return Err(ContractError::InsufficientAmount {
                     denom: fee_plus_tip.denom,
                     requested: fee_plus_tip.amount,
-                    provided: found_coin_amount,
+                    provided: fund_coin_amount,
                 });
             }
         }
@@ -180,24 +180,24 @@ mod tests {
     }
 
     #[test]
-    fn fixed_fees_insufficient_founds() {
+    fn fixed_fees_insufficient_funds() {
         let fixed_fee_amount = 2000;
         let tip_amount = 1000;
-        let found_amount = 2500;
+        let fund_amount = 2500;
 
         let service_fees = StateServiceFee::Fixed {
             amount: vec![Coin::new(fixed_fee_amount, "udsm")],
         };
 
-        let founds = vec![Coin::new(found_amount, "udsm")];
+        let funds = vec![Coin::new(fund_amount, "udsm")];
         let tips = vec![Coin::new(tip_amount, "udsm")];
-        let computed_fees = service_fees.check_fees(&founds, &tips).unwrap_err();
+        let computed_fees = service_fees.check_fees(&funds, &tips).unwrap_err();
 
         assert_eq!(
             ContractError::InsufficientAmount {
                 denom: "udsm".to_string(),
                 requested: Uint128::new(fixed_fee_amount + tip_amount),
-                provided: Uint128::new(found_amount),
+                provided: Uint128::new(fund_amount),
             },
             computed_fees
         );
@@ -233,7 +233,7 @@ mod tests {
     }
 
     #[test]
-    fn fixed_fees_fee_found_coin_not_provided() {
+    fn fixed_fees_fee_fund_coin_not_provided() {
         let fixed_fee_amount = 20000;
         let tip_amount = 100000;
 
@@ -313,9 +313,9 @@ mod tests {
     }
 
     #[test]
-    fn percentage_fees_insufficient_founds() {
+    fn percentage_fees_insufficient_funds() {
         let tip_amount = 1000;
-        let found_amount = 1099;
+        let fund_amount = 1099;
 
         // Fee at 10%
         let service_fees = StateServiceFee::Percentage {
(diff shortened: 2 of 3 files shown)


contracts/tips/src/utils.rs
```diff
@@ -0,0 +1,75 @@
+use cosmwasm_std::Coin;
+use std::collections::btree_map::BTreeMap;
+
+/// Iterates over the coins vector and merge the coins having the same `denom.
```
bragaz, line 4:
```suggestion
/// Iterates over the coins vector and merges the coins having the same `denom`.
```

bragaz (commented):
Awesome job, I've made some minor comments but nothing code-related.

bragaz:
I would also like @RiccardoM to give a look at it if he has some time

manu0466:
I add Riccardo as reviewer.
@bragaz I have also made some minor changes to make the `ServiceFee` Optional to allow the contract to not take any fees from the tipped amount in a more clear way insted of using a 0 `ServiceFee:.Percentage` or an empty `ServiceFee::Fixed`

contracts/tips/src/msg.rs
```diff
@@ -0,0 +1,282 @@
...
+use crate::state::StateServiceFee;
+use cosmwasm_std::{Addr, Coin, Uint128, Uint64};
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+/// Fees required to execute [`ExecuteMsg::SendTip`].
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum ServiceFee {
+    /// Represents a fixed fee that is deducted from the tip.
```
RiccardoM, line 11: I see that both the `Fixed` and `Percentage` type of fees state the following: 
> [...] fee that is deducted from the tip.

From a user perspective, it looks like if the user wants to send 1 DSM and the fee is 0.01 DSM: 
- the sender will end up paying 1 DSM 
- the receiver will receive 0.99 DSM (1 DSM - 0.01 DSM)

Although this is works, I think that instead the fee amount should be payed by the sender, and not subtracted from the tip amount. So, if a sender wants to send a 1 DSM tip and the fee is 0.01 DSM the following should happen: 
- the sender pays 1.01 DSM 
- the receiver pays 0.99 DSM

This will also make it easier to be displayed in a UI and will cause less confusion from the sender's perspective. 

If this is already the case of how things work, I think this comment should be edited into something like
> [...] fee that the sender will have to pay.

contracts/tips/src/msg.rs
```diff
@@ -0,0 +1,282 @@
...
+        /// Percentage decimals.
+        /// Example if this value is 3 means that the value should have 3 decimals numbers so
+        /// if value is
+        /// - 1 means 0,001%
+        /// - 10 means 0,01%
+        /// - 100 means 0,1%
+        /// - 1000 means 1%
+        /// - 10000 means 10%
+        /// - 100000 means 100%
+        decimals: u32,
```
RiccardoM, line 29: This is extremely confusing to read and understand. I think the overall explanation can be improved somehow to make it more clear

contracts/tips/src/msg.rs
manu0466: In the example of the desired behaviour you mean
* the sender pays 1.01 DSM
* the receiver receives 1 DSM  

If so the desired behaviour that you are describing is already implemented for the `Fixed` but not for `Percentage`.
I can extend the `SendTip` message to have a `amount` field that represents the amount that the sender want tip and calculate the percentage from this `amount` field, then if `calculated fee` + `amount` is equal to the coins received in the `MessageInfo::amount` we consider the tip valid.  

cc @bragaz

contracts/tips/src/utils.rs
```diff
@@ -0,0 +1,75 @@
...
+use std::collections::btree_map::BTreeMap;
+
+/// Iterates over the coins vector and merges the coins having the same `denom`.
+/// * `coins` - Vector of coins to merge.
+pub fn merge_coins(coins: Vec<Coin>) -> Vec<Coin> {
+    if coins.len() <= 1 {
+        return coins;
+    }
+
+    let mut map: BTreeMap<String, u128> = BTreeMap::new();
```
dadamu, line 11: Can we use HashMap, instead?
BTreeMap is for the situation when the keys need to be sorted, see [reference](https://doc.rust-lang.org/std/collections/index.html#use-a-btreemap-when).

contracts/tips/src/utils.rs
```diff
@@ -0,0 +1,75 @@
...
+    for coin in coins {
+        let value = map.get_mut(&coin.denom);
+
+        if let Some(amount) = value {
+            *amount += coin.amount.u128();
```
dadamu, line 16: Should we use the safe math Cosmwasm provide instead the raw add?
https://docs.rs/cosmwasm-std/0.14.0-rc1/cosmwasm_std/struct.Uint128.html#method.checked_add

contracts/tips/src/msg.rs
```diff
@@ -0,0 +1,282 @@
...
+pub enum ServiceFee {
+    /// Represents a fixed fee that is deducted from the tip.
+    Fixed {
+        /// Coins that are deducted from the tip.
+        amount: Vec<Coin>,
+    },
+    /// Represents a percentage that is deducted from the tip.
+    Percentage {
+        /// Percentage value.
+        value: Uint128,
```
dadamu, line 19: I think we can use [`Decimal`]([from_ratio](https://docs.rs/cosmwasm-std/0.9.2/cosmwasm_std/struct.Decimal.html#method.from_ratio)), instead.

contracts/tips/src/contract.rs
```diff
@@ -0,0 +1,1728 @@
...
+            // Add the new key to the end
+            tips_key_list.push_back(tip_key);
+            TIPS_KEY_LIST.save(deps.storage, &tips_key_list)?;
+            // Return the new coins
+            coin_to_send.clone()
+        };
+        tip_record_key.save(deps.storage, &tip_record_coins)?;
+    }
+
+    if coin_to_send.is_empty() {
```
dadamu, line 161: It can be moved before the line 138 `if config.tips_record_threshold > 0` to check the funds is not zero first.

contracts/tips/src/msg.rs
```diff
@@ -0,0 +1,266 @@
...
+pub struct InstantiateMsg {
+    /// Address of who will have the right to manage the contract.
+    pub admin: String,
+    /// Application which is deploying the contract.
+    pub subspace_id: Uint64,
+    /// Fee that the users need to pay to use the contract.
+    /// If `None` no fees will be collected from the tipped amount.
+    pub service_fee: Option<ServiceFee>,
+    /// The number of records saved of a user tips history.
+    pub saved_tips_threshold: u32,
```
dadamu, line 69: I would name it `saved_tips_size`. Threshold to me is something that the lowest condition to save.

manu0466:
The contract has changed a little bit after talking with @bragaz regarding how the tip history should be kept.
Before the contract was using the `threshold` value to limit the total number of tips memorized in the contract state, now that value is used to limit the history of tips sent and received from a user. 
Please @bragaz take another look at it.

Also since now i keep the `block height` to sort the tips from the oldest to the latest we can return that information when querying the tips what do you think guys?

contracts/tips/src/msg.rs
```diff
@@ -0,0 +1,278 @@
...
```
dadamu, line 69:
```suggestion
    pub saved_tips_record_size: u32,
```

contracts/tips/src/contract.rs
```diff
@@ -0,0 +1,2044 @@
...
+// version info for migration info
+const CONTRACT_NAME: &str = "crates.io:tips";
+const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
+
+// actions for executing messages
+const ACTION_INSTANTIATE: &str = "instantiate";
+const ACTION_SEND_TIP: &str = "send_tip";
+const ACTION_UPDATE_SERVICE_FEE: &str = "update_service_fee";
+const ACTION_UPDATE_ADMIN: &str = "update_admin";
+const ACTION_UPDATE_SAVED_TIPS_RECORD_THRESHOLD: &str = "update_saved_tips_record_threshold";
```
dadamu, line 29: Can we rename all `record_threshold` into `record_size`?

contracts/tips/src/contract.rs
```diff
@@ -0,0 +1,2044 @@
...
+            Some(ServiceFee::Fixed { amount: vec![] }),
+            1,
+        )
+        .unwrap_err();
+        assert_eq!(ContractError::InvalidSubspaceId {}, init_err);
+    }
+
+    #[test]
+    fn init_contract_with_non_existing_subspace() {
+        let querier = MockQuerier::<DesmosQuery>::new(&[(MOCK_CONTRACT_ADDR, &[])])
```
dadamu, line 567: Great method to mock deps. I think we can have this method in desmos-bindings allows devs to mock what they need instead of the default.

manu0466: Yes! was thikning the same, we can have something like `MockDesmosQuerier::new(...)` and with methods to mock the response of each module like `with_custom_subspaces_handler` or `with_custom_profiles_handler`, what you think?

contracts/tips/src/state.rs
```diff
@@ -0,0 +1,365 @@
...
+        },
+    )
+}
+
+impl StateServiceFee {
+    /// Computes the fees that the contract will holds and the coins that
+    /// can be sent to the user.
+    /// * `founds` - Coins sent from the user to the contract.
+    /// * `tip_amount` - Coins from which to calculate the fees.
+    pub fn check_fees(&self, founds: &[Coin], tip_amount: &[Coin]) -> Result<(), ContractError> {
```
dadamu, line 96:
```suggestion
    pub fn check_fees(&self, funds: &[Coin], tip_amount: &[Coin]) -> Result<(), ContractError> {
```

contracts/tips/src/state.rs
```diff
@@ -0,0 +1,365 @@
...
+            ),
+            post_id: MultiIndex::new(|tip| tip.post_id, "tips_record", "tips_record__post_id"),
+        },
+    )
+}
+
+impl StateServiceFee {
+    /// Computes the fees that the contract will holds and the coins that
+    /// can be sent to the user.
+    /// * `founds` - Coins sent from the user to the contract.
```
dadamu, line 94:
```suggestion
    /// * `funds` - Coins sent from the user to the contract.
```

contracts/tips/src/state.rs
```diff
@@ -0,0 +1,365 @@
...
+                assert_eq!(fees, amount)
+            }
+            StateServiceFee::Percentage { .. } => {
+                panic!("ServiceFee::Fixed should be converted to StateServiceFee::Fixed")
+            }
+        }
+    }
+
+    #[test]
+    fn fixed_fees_insufficient_founds() {
```
dadamu, line 183:
```suggestion
    fn fixed_fees_insufficient_funds() {
```

contracts/tips/src/state.rs
```diff
@@ -0,0 +1,365 @@
...
+        let fixed_fee_amount = 2000;
+        let tip_amount = 1000;
+        let found_amount = 2500;
```
dadamu, line 186:
```suggestion
        let fund_amount = 2500;
```

contracts/tips/src/state.rs
```diff
@@ -0,0 +1,365 @@
...
+
+        let service_fees = StateServiceFee::Fixed {
+            amount: vec![Coin::new(fixed_fee_amount, "udsm")],
+        };
+
+        let founds = vec![Coin::new(found_amount, "udsm")];
```
dadamu, line 192:
```suggestion
        let funds = vec![Coin::new(found_amount, "udsm")];
```

contracts/tips/src/state.rs
```diff
@@ -0,0 +1,365 @@
...
+                denom: "uatom".to_string(),
+                requested: fixed_fee_amount.into(),
+                provided: Uint128::zero()
+            },
+            computed_fees
+        );
+    }
+
+    #[test]
+    fn fixed_fees_fee_found_coin_not_provided() {
```
dadamu, line 236:
```suggestion
    fn fixed_fees_fee_fund_coin_not_provided() {
```

contracts/tips/src/state.rs
```diff
@@ -0,0 +1,365 @@
...
+                panic!("ServiceFee::Percentage should be converted to StateServiceFee::Percentage")
+            }
+            StateServiceFee::Percentage { value } => {
+                assert_eq!("1", value.to_string());
+            }
+        }
+    }
+
+    #[test]
+    fn percentage_fees_insufficient_founds() {
```
dadamu, line 316:
```suggestion
    fn percentage_fees_insufficient_funds() {
```

contracts/tips/src/state.rs
```diff
@@ -0,0 +1,365 @@
...
+        let tip_amount = 1000;
+        let found_amount = 1099;
+
+        // Fee at 10%
+        let service_fees = StateServiceFee::Percentage {
+            value: Decimal::from_atomics(10u128, 0u32).unwrap(),
+        };
+
+        let tips = vec![Coin::new(tip_amount, "udsm")];
+        let founds = vec![Coin::new(found_amount, "udsm")];
```
dadamu, line 326:
```suggestion
        let funds = vec![Coin::new(fund_amount, "udsm")];
```

contracts/tips/src/state.rs
```diff
@@ -0,0 +1,365 @@
...
+            StateServiceFee::Percentage { value } => {
+                assert_eq!("1", value.to_string());
+            }
+        }
+    }
+
+    #[test]
+    fn percentage_fees_insufficient_founds() {
+        let tip_amount = 1000;
+        let found_amount = 1099;
```
dadamu, line 318:
```suggestion
        let fund_amount = 1099;
```

dadamu:
There are typos that `found` should be fund.

manu0466:
After having a call with @bragaz we need to change some behaviour of this contract and the ADR-004. I close this PR for now

PR #59, opened 2022-09-12
feat: ADR-003 remarkables contract implementation

dadamu:
This PR implements remarkables contract and update ADR of it.


contracts/remarkables/src/contract.rs
```diff
@@ -0,0 +1,699 @@
...
+        .add_message(wasm_execute_mint_msg))
+}
+
+fn is_enough_fees(funds: Vec<Coin>, requireds: Vec<Coin>) -> bool {
+    if requireds.len() == 0 {
+        return true;
+    }
+    // It takes O(n^2) time complexity but both list are extremely small
+    for required in requireds.iter() {
+        if has_coins(&funds, &required) {
```
manu0466, line 177: I think that this if should be `!has_coins` and should return `false` and return `true` at line 178 otherwise if we have two lists of coins like this:
1. founds: [100udsm]
2. requireds: [100uds, 100uatom]  

This function will wrongly return true.
Please also add a test with such case to have it covered.

contracts/remarkables/src/contract.rs
```diff
@@ -0,0 +1,699 @@
...
+        subspace_id: config.subspace_id.into(),
+    })
+}
+
+fn query_rarities(deps: Deps<DesmosQuery>) -> StdResult<QueryRaritiesResponse> {
+    let res: StdResult<Vec<_>> = RARITY
+        .range(deps.storage, None, None, Order::Ascending)
+        .collect();
+    let rarities = res?
+        .iter()
```
manu0466, line 299: Use [Vec::drain](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.drain) insetad to avoid the copy of `level` and `mint_fees` inside `map`.

contracts/remarkables/src/state.rs
```diff
@@ -0,0 +1,23 @@
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+use cosmwasm_std::{Addr, Coin};
+use cw_storage_plus::{Item, Map};
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct RarityState {
+    pub level: u32,
```
manu0466, line 9: Can we remove this? since is already in the key of `RARITY`

contracts/remarkables/src/state.rs
```diff
@@ -0,0 +1,23 @@
...
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct ConfigState {
+    pub admin: Addr,
+    pub subspace_id: u64,
+    pub cw721_code_id: u64,
+}
+
+pub const RARITY: Map<u32, RarityState> = Map::new("rarity");
```
manu0466, line 21: Can we use a `Vec` instead? If we remove the `level` from `msg::Rarity` and use the index of `InstantiateMsg::rarities` as rarity level we don't need a Map and we can access the rarity levels in O(1)

contracts/remarkables/src/contract.rs
dadamu: Thanks for spotting the issue, I thought required fees should be `or` condition.

dadamu: Thanks for spotting the issue. I thought required fees should be `or` condition, instead of `and`. I would like to confirm which condition in mint fees should be.
cc @bragaz

contracts/remarkables/src/state.rs
dadamu: I have same thought to use `Vec` instead of `Map`. Finally, I decided to use `Map` since It forces the rarity level must be sequential and start from 0, and `Map` provides more flexibility if we need to add more rarity level in the future.
But yeah, it makes more sense to use vector because I can't find the use case if the current contract has no needs to add more levels and customize the level number.

contracts/remarkables/src/contract.rs
```diff
@@ -0,0 +1,699 @@
...
+    check_eligibility(
+        deps.storage,
+        deps.querier.deref(),
+        info.sender.clone(),
+        post_id,
+        rarity.engagement_threshold,
+    )?;
+    // Create the cw721 message to send to mint the remarkables
+    let mint_msg = Cw721ExecuteMsg::<Empty, Empty>::Mint(MintMsg::<Empty> {
+        token_id: post_id.to_string(),
```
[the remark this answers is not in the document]
dadamu: Implemented the possibility to mint different level on a post.

contracts/remarkables/Cargo.toml
```diff
@@ -0,0 +1,48 @@
+[package]
+name = "remarkables"
+version = "0.1.0"
+authors = ["Paul <p22626262@gmail.com>"]
```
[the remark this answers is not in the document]
dadamu: Changed.

contracts/remarkables/src/contract.rs
```diff
@@ -0,0 +1,922 @@
...
+            key: None,
+            offset: None,
+            limit: Uint64::new(1),
+            count_total: true,
+            reverse: false,
+        }),
+    )?;
+    let reactions_pagination = reactions.pagination.unwrap_or(PageResponse::default());
+    let reactions_count = reactions_pagination.total.unwrap_or(0u64.into());
+    if engagement_threshold as u64 > reactions_count.into() {
```
manu0466, line 228: Should we remove the reactions made from the sender from the reaction count? 
This to prevent that a user cheat on the number of reactions of a post

contracts/remarkables/src/contract.rs
```diff
@@ -0,0 +1,922 @@
...
+
+fn execute_update_admin(
+    deps: DepsMut<DesmosQuery>,
+    info: MessageInfo,
+    new_admin: String,
+) -> Result<Response<DesmosMsg>, ContractError> {
+    check_admin(deps.storage, &info)?;
+    let new_admin_addr = deps.api.addr_validate(&new_admin)?;
+    CONFIG.update(deps.storage, |mut config| -> Result<_, ContractError> {
+        config.admin = new_admin_addr.clone();
```
manu0466, line 242: use new
```suggestion
        config.admin = new_admin_addr;
```
Just take the ownership here and use `new_admin` inside the response to avoid a copy

contracts/remarkables/src/contract.rs
```diff
@@ -0,0 +1,922 @@
...
+        Ok(config)
+    })?;
+    Ok(Response::new()
+        .add_attribute(ATTRIBUTE_ACTION, ACTION_UPDATE_ADMIN)
+        .add_attribute(ATTRIBUTE_SENDER, &info.sender)
+        .add_attribute(ATTRIBUTE_ADMIN, &info.sender)
+        .add_attribute(ATTRIBUTE_NEW_ADMIN, new_admin_addr))
```
manu0466, line 249:
```suggestion
        .add_attribute(ATTRIBUTE_NEW_ADMIN, new_admin))
```

contracts/remarkables/src/msg.rs
```diff
@@ -0,0 +1,241 @@
...
+    pub subspace_id: Uint64,
+    /// List of rarities to initialize with this contract.
+    pub rarities: Vec<Rarity>,
+}
+
+impl InstantiateMsg {
+    /// Checks that the data inside the message are coherent.
+    /// NOTE: This function don't checks if the address are valid.
+    pub fn validate(&self) -> Result<(), ContractError> {
+        if self.subspace_id.is_zero() {
```
manu0466, line 26: Can youa also do the same check for `cw721_code_id` since is > 0

manu0466 (commented):
Just some minor issues, overall the contract looks good to me.  
Let's wait for @bragaz feedback on the other doubths.

docs/architecture/adr-003-remarkables-contract.md
```diff
@@ -145,18 +145,27 @@ pub struct QueryConfigResponse {
...
-The `RarityMintFees` query return the rarity mint fees required to mint a specific Remarkable inside a `RarityMintFeesResponse`.
+### Tokens
+This query returns all the ids of tokens owned by the given owner from cw721 contract.
 ```rust
-pub struct QueryRarityMintFeesResponse {
-  pub rarity_mint_fees: Vec<RarityMintFee>
+pub struct TokensResponse {
+    pub tokens: Vec<String>,
 }
 ```
```
bragaz, line 171: About the `extention` field, I think we can use it to add some meaningful metadata that should become useful for the use cases I envision for the remarkables. In particular, we should have such `Metadata` on chain:
```rust
pub struct Metadata {
pub rarity_level: u32,
pub subspace_id: u64,
}

```

contracts/remarkables/src/contract.rs
```diff
@@ -0,0 +1,922 @@
...
+        .add_attribute(ATTRIBUTE_RARITY_LEVEL, rarity_level.to_string())
+        .add_attribute(ATTRIBUTE_TOKEN_ID, token_id.to_string())
+        .add_attribute(ATTRIBUTE_RECIPIENT, &info.sender)
+        .add_attribute(ATTRIBUTE_TOKEN_URI, remarkables_uri)
+        .add_message(wasm_execute_mint_msg))
+}
+
+// Returns the token id as "<post-id>-<rarity-level>"
+pub fn convert_post_id_to_token_id(post_id: u64, rarity_level: u32) -> String {
+    post_id.to_string() + "-" + &rarity_level.to_string()
```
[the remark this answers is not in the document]
manu0466: @bragaz I think that is not necessary, since for each instance of `remarkable` contract there's a different instance of `cw721` so such collision should never occur.
Also in the current implementation we don't allow to update the `cw721` address to one associated with another another instance of `remarkable`.

contracts/remarkables/src/contract.rs
dadamu: Fixed

contracts/remarkables/src/state.rs
dadamu: Cool, changed to `Vec`.

contracts/remarkables/src/contract.rs
dadamu: Fixed

contracts/remarkables/src/msg.rs
dadamu: Added

contracts/remarkables/src/contract.rs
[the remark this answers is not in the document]
dadamu: Yes, agree with @manu0466, the collision would never occur since a `cw721` only belongs one `remarkable` contract.

contracts/remarkables/src/msg.rs
```diff
@@ -0,0 +1,241 @@
...
+    pub engagement_threshold: u32,
+    /// Minimum fees of minting.
+    pub mint_fees: Vec<Coin>,
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum ExecuteMsg {
+    /// Message allowing the user to mint a Remarkables for a specific post owned by the user.
+    MintTo {
```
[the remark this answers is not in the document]
dadamu: Updated

docs/architecture/adr-003-remarkables-contract.md
dadamu: The level and post id is including in the token id. If the client want to know the rarity level or post id from a token, they can parse token id, besides, subspace id can be queried from the remarkables config. From my side, `Metadata` seems duplicated.

contracts/remarkables/src/contract.rs
```diff
@@ -0,0 +1,997 @@
...
+                count_total: true,
+                reverse: false,
+            }),
+        )?
+        .pagination
+        .unwrap_or(PageResponse::default())
+        .total
+        .unwrap_or(0u64.into());
+    let a = total_reactions_count.checked_sub(self_reactions_count)?;
+    println!("{}", a);
```
manu0466, line 248: Please remove the `println!`

dadamu: Thank you, forgot to remove. Removed!

manu0466 (approved):
Looks good to me, nice job :muscle:

contracts/remarkables/src/contract.rs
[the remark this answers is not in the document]
bragaz: Ok cool!

docs/architecture/adr-003-remarkables-contract.md
dadamu: Yeah, reading from `Metadata` field would cost less. consider the communication among each contracts, having a `Metadata` would be useful.

dadamu: Added

contracts/cw721-remarkables/examples/schema.rs
```diff
@@ -0,0 +1,25 @@
...
+    export_schema_with_title(
+        &schema_for!(AllNftInfoResponse<Metadata>),
+        &out_dir,
+        "AllNftInfoResponse",
+    );
+    export_schema_with_title(
+        &schema_for!(NftInfoResponse<Metadata>),
+        &out_dir,
+        "NftInfoResponse",
+    );
```
dadamu, line 24: Since this cw721 contract only has the difference about Metadata from cw721-base, we just need to expose `AllNftInfoResponse<Metadata>` and `NftInfoResponse<Metadata>`. The rest can be obtained from cw721-base.

dadamu (commented):
Follow the guide to add error or properly in the end of unit test function name, please.

contracts/remarkables/src/contract.rs
```diff
@@ -0,0 +1,1000 @@
...
+    // It takes O(n^2) time complexity but both list are extremely small
+    for required in requireds.iter() {
+        if !has_coins(&funds, &required) {
+            return false;
+        }
+    }
+    true
+}
+
+/// Checks the post reaches the engagement threshold.
```
bragaz, line 197:
```suggestion
/// Checks that the post reaches the engagement threshold.
```

contracts/remarkables/src/contract.rs
```diff
@@ -0,0 +1,1000 @@
...
+    Ok(Response::new()
+        .add_attribute(ATTRIBUTE_ACTION, ACTION_INSTANTIATE)
+        .add_attribute(ATTRIBUTE_SENDER, &info.sender)
+        .add_attribute(ATTRIBUTE_ADMIN, msg.admin)
+        .add_attribute(ATTRIBUTE_CW721_CODE_ID, msg.cw721_code_id)
+        .add_submessage(cw721_submessage))
+}
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn execute(
```
bragaz, line 109: @dadamu Since I would like to make possible in the future to burn Remarkables to obtain in-app advantages, could you check if, once burned, the token-id can be re-minted? I've checked the `Burn` method and it looks like it's not possible but I would like to have your confirmation on this

dadamu: It is possible, cw721 `Burn` method removed the token-id in its state.

dadamu: Disallowed the behavior to mint burned token post again.

bragaz (approved):
Awesome job 🚀

PR #70, opened 2022-09-14
build(ci): run lint tests only if a .rs file is changed

manu0466:
This PR updates the CI workflow to run the lint check only if a `.rs` file is changed in order to reduce the amount of time required to performs the automatic checks

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -15,14 +15,14 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Verify .rs files 👀
-        uses: tj-actions/verify-changed-files@v11.1
-        id: verify-changed-files
+        uses: technote-space/get-diff-action@v6.1.0
+        id: git_diff
         with:
-          files: |
-            **/*.rs
+          PATTERNS: |
+            **/**.rs
 
       - name: Setup Rust ⚙
-        if: steps.verify-changed-files.outputs.files_changed == 'true'
+        if: env.GIT_DIFF
         uses: actions-rs/toolchain@v1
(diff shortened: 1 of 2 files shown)


dadamu (approved):

.github/workflows/lint.yml
```diff
@@ -14,7 +14,15 @@ jobs:
       - name: Checkout 🛎️
         uses: actions/checkout@v2
 
-      - name: Setup Rust ⚙️
+      - name: Verify .rs files 👀
+        uses: tj-actions/verify-changed-files@v11.1
```
RiccardoM, line 18: Why are we using this actions instead of [technote-space/get-diff-action](https://github.com/technote-space/get-diff-action) like we do on other repos as well (i.e. [Desmos](https://github.com/desmos-labs/desmos/blob/master/.github/workflows/test.yml#L54))?

I would prefer if we can stick to the same action throught all the repos if possible, so it's easier to manage for everyone.

.github/workflows/test.yml
```diff
@@ -20,22 +20,7 @@ jobs:
...
-
-      - name: Test POAP contract 🧪
-        run: cargo test --package poap
-        env:
-          RUST_BACKTRACE: 1
-
-      - name: Test POAP manager contract 🧪
-        run: cargo test --package poap-manager
+      - name: Test contracts 🧪
+        run: cargo test
```
RiccardoM, line 24: We should run the tests only if `.rs` files (other other noteworthy file types) have changed as well

RiccardoM (approved):

Issue #189, opened 2023-07-11
POAP: Potential Misnaming of the 'MINTER_ADDRESS' State Variable


giorgionocera:
## Feature description
The state variable `MINTER_ADDRESS` is currently used to store data related to the tokens (POAPs) minted to different wallets (addresses). Here, the address serves as the key, and the value represents the number of POAPs minted to that address https://github.com/desmos-labs/desmos-contracts/blob/946688ef0d0671cd1497d5f7baf361587baac6c7/contracts/poap/src/state.rs#L33 Given its usage, the name "MINTER_ADDRESS" might not accurately reflect the data it holds.

## Implementation proposal
Considering the data held by this state variable, a more descriptive name could be `MINTED_TOKENS`. This change would make the purpose of this variable clearer to developers who read the code, improving the readability and understandability of the codebase.

Issue #188, opened 2023-07-11
POAP: Possible Oversight in README Regarding MintTo Function


giorgionocera:
## Feature description
It seems there might be a discrepancy in the [poap contract README](https://github.com/desmos-labs/desmos-contracts/blob/946688ef0d0671cd1497d5f7baf361587baac6c7/contracts/poap/README.md) specifically in the description of the MintTo function https://github.com/desmos-labs/desmos-contracts/blob/946688ef0d0671cd1497d5f7baf361587baac6c7/contracts/poap/README.md?plain=1#L76-L77 In the code, MintTo calls the `execute_mint()` function, passing the `check_authorized_to_mint` parameter as `true`. Therefore, it performs the following check:
https://github.com/desmos-labs/desmos-contracts/blob/946688ef0d0671cd1497d5f7baf361587baac6c7/contracts/poap/src/contract.rs#L216-L219
Thus, both the admin and the minter can mint a token, contradicting the README, which states that "the minter can mint a POAP token if the contract enables mint". Moreover, the `minter` and the `admin` are able to mint the token even if the "mint is not enabled".
https://github.com/desmos-labs/desmos-contracts/blob/946688ef0d0671cd1497d5f7baf361587baac6c7/contracts/poap/src/contract.rs#L211-L214

## Implementation proposal
The README should correctly reflect the code's functionality. In this case, it should clarify that both the admin and the minter are able to mint a token and that they bypass the `mint_enabled` check.

Issue #187, opened 2023-07-11
POAP: Duplication of Checks in mint and mintto Functions


giorgionocera:
## Feature description
Currently, there are identical checks being performed in both the `ExecuteMsg:Mint` https://github.com/desmos-labs/desmos-contracts/blob/946688ef0d0671cd1497d5f7baf361587baac6c7/contracts/poap/src/contract.rs#L139-L142 and `ExecuteMsg:MintTo` https://github.com/desmos-labs/desmos-contracts/blob/946688ef0d0671cd1497d5f7baf361587baac6c7/contracts/poap/src/contract.rs#L143-L146 which use the same function `execute_mint()`. This raises a question: why are these checks duplicated, instead of being centralized within the `execute_mint()` function, similar to how the check for the new admin's address is handled in the `execute_update_admin()` https://github.com/desmos-labs/desmos-contracts/blob/946688ef0d0671cd1497d5f7baf361587baac6c7/contracts/poap/src/contract.rs#L331 function?

## Implementation proposal
Even if it is not a huge improvement, if there is no specific reason for this duplication, it might be more efficient to consolidate these checks within the `execute_mint` function. This could enhance code readability and maintainability by avoiding unnecessary repetition.

Issue #186, opened 2023-07-11
POAP: Considering Use of CW721-non-transferable Instead of CW721-base


giorgionocera:
## Feature description
The current implementation uses a wrapper of CW721-base
https://github.com/desmos-labs/desmos-contracts/blob/946688ef0d0671cd1497d5f7baf361587baac6c7/contracts/cw721-poap/src/lib.rs#L11-L14 storing the address of the claimer to whom the POAP is minted
https://github.com/desmos-labs/desmos-contracts/blob/946688ef0d0671cd1497d5f7baf361587baac6c7/contracts/cw721-poap/src/lib.rs#L7-L9 However, the NFT remains transferable in this setup. It might be worthwhile to consider whether the implementation should omit the claimer's address and instead set the NFT as [non-transferable](https://github.com/CosmWasm/cw-nfts/tree/main/contracts/cw721-non-transferable), binding it indelibly to that address.

## Implementation proposal
It would be beneficial to have a discussion on whether the approach of using a CW721-base with claimer information is more advantageous compared to using a CW721-non-transferable for the use case of POAP NFTs. If there's solid reasoning behind the existing design, it should be documented for better understanding. If not, and if the CW721-non-transferable is a more suitable choice, then perhaps the implementation could be reconsidered.

PR #195, opened 2023-07-17
feat(ADR-006): POAP v2

manu0466:
## Description

This PR updates the POAP contract to make it compliant with the ADR-006.

Closes: #182
Closes: #183
Closes: #184 
Closes: #185 
Closes: #186 
Closes: #187 
Closes: #188 
Closes: #189

---

### Author Checklist

*All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues.*

I have...

- [x] included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] added `!` to the type prefix if API or client breaking change
- [x] targeted the correct branch (see [PR Targeting](https://github.com/desmos-labs/desmos-contracts/blob/master/CONTRIBUTING.md#pr-targeting))
- [ ] provided a link to the relevant issue or specification
- [x] included the necessary unit and integration [tests](https://github.com/desmos-labs/desmos-contracts/blob/master/CONTRIBUTING.md#testing)
- [ ] added a changelog entry to `CHANGELOG.md`
- [ ] updated the relevant documentation or specification
- [ ] reviewed "Files changed" and left comments if necessary
- [ ] confirmed all CI checks have passed

### Reviewers Checklist

*All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items.*

I have...

- [ ] confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] confirmed `!` in the type prefix if API or client breaking change
- [ ] confirmed all author checklist items have been addressed
- [ ] reviewed contract logic
- [ ] reviewed contract security
- [ ] reviewed API design and naming
- [ ] reviewed documentation is accurate
- [ ] reviewed tests and test coverage
- [ ] manually tested (if applicable)


contracts/poap/src/execute.rs
```diff
@@ -0,0 +1,409 @@
...
+    /// * `sender` - Address that will be checked.
+    pub fn assert_is_minter(
+        &self,
+        storage: &dyn Storage,
+        sender: &Addr,
+    ) -> Result<(), ContractError> {
+        let minter = self.minter.load(storage)?;
+        if !sender.eq(&minter) {
+            return Err(ContractError::MintUnauthorized {});
+        }
```
RiccardoM, line 357: We need to allow the contract admin to act as a minter. If the `minter` is not set, nobody, including the admin, will be able to mint tokens. On the contrary, with this new check, if the `minter` is set, only that specified person can mint tokens. To ensure this functionality works correctly, we should add tests for both cases:
1. When the `minter` is specified, the admin can mint some tokens.
2. When the `minter` is not specified, the admin can still mint some tokens.

giorgionocera:
Hello,

First of all, thank you for providing this codebase.

I've noticed that there might be a lack of validation for the minting date range. Currently, there doesn't seem to be a check that ensures the `end date` for minting is later than the `start date`.

For example, when modifying a test (like the one shown here: https://github.com/desmos-labs/desmos-contracts/blob/2f6629455090a1418fb5e77ffef5610c0edf50a1/contracts/poap/src/contract_tests.rs#L139) with the following values:

```rust
mint_start_time: Some(Timestamp::from_seconds(10)), 
mint_end_time: Some(Timestamp::from_seconds(8))
```
the test does not fail, which it arguably should because it's impossible for the end time to be before the start time.

I believe adding validation to check that the end date is always greater than the start date could help prevent potential issues in the future.

Thank you for your consideration.

giorgionocera:
Hello,

Thank you for your work on this codebase. 

I would like to propose a change regarding the handling of the `minter` field.

In certain scenarios, it could be desirable to have the ability to set the `minter` to `None`. For instance, if one wants to ensure that only the admin can mint tokens, it might make sense to not set a third-party minter.

Currently, the `update_minter` function and `UpdateMinter` message does not accept `None` as a valid value for the `minter`. I propose that this be changed to accept `None` values.

This change could include:
- Modifying the `update_minter` function to accept `None` as a valid value for `minter`.
- Updating the `UpdateMinter` message to allow for `None` as a valid `minter` value.
  
In addition, in this scenario, we might consider allowing `None` as a minter during the contract instantiation phase.

This change could provide additional flexibility for the management of minting permissions.

Thank you for considering this proposal.

giorgionocera:
I also noticed that there are currently no checks in place to ensure that the `mint_start_time` and `mint_end_time` provided are greater than the current time.

A future timestamp for both `mint_start_time` and `mint_end_time` should be used to ensure logical consistency in the token minting process. Moreover, it seems logical that a token-minting operation should occur in the future, not in the past.

I propose adding a validation step to the relevant functions to check that both `mint_start_time` and `mint_end_time` are strictly greater than the current time.

giorgionocera:
While the current implementation is different, in the context of a "proof of participation" system, it would be logical for each user to hold a maximum of one token per event, indicating whether they have participated or not as explained here: https://github.com/desmos-labs/desmos-contracts/pull/190#discussion_r1260969108. If we aim for a more generic implementation where users can hold multiple tokens, we might consider implementing a `poap_mint_limit_per_address` variable, which specifies the maximum number of tokens that can be minted per address.
What do you think about this?
Thank you!

manu0466:
@RiccardoM @giorgionocera I update the contract adding the following features:
1. Allow the `admin` to perform the same operations as the `minter`;
2. Allow the `minter` to be `None`;
3. Limit the POAP that an user can mint to 1.

As for the checks to make sure that the `mint_start_time` and `mint_end_time` are in the future, I think they are not necessary, also not having such checks can avoid problems. For example, suppose you want to instantiate the contract with `mint_start_time` in the next 10 seconds, if the message is not included in the next 2 blocks the transaction will fail, because 12 seconds have elapsed from when the transaction was transmitted to when it is included.

contracts/poap/src/execute.rs
```diff
@@ -392,6 +392,18 @@ where
             return Err(ContractError::MintDisabled {});
         }
 
+        // Check if this user have already minted a poap.
```
giorgionocera, line 395: By putting this check here, the admin and the minter are still able to mint more than a poap to each user. Is this behavior a desired one?

manu0466: Since I did not have any specifics on this behavior, I decided to give them the option of minting as many poaps as they want.  In any case, minting can be limited for them as well.
Keep in mind that even with the `MintTo` action it is possible to mint more than one POAP for a user, in this case, however, I would like to avoid checking if the user already has a POAP to save gas, since this is an operation performed by a trusted party.

RiccardoM: @manu0466 I believe we can consistently check that each user has only one POAP minted. As this is a **read** operation, it won't cost gas (gas is only used for **write** operations). This ensures that the behavior remains consistent.

RiccardoM (commented):
Overall looks good to me. However, it seems to me that there are some duplicated tests. Particularly: 

- `admin_can_mint_before_start_time` and `admin_can_mint_after_end_time` are duplicates of `admin_can_mint_outside_mint_time`
- `minter_can_mint_before_start_time` and `minter_can_mint_after_end_time` are duplicates of `minter_can_mint_outside_mint_time`
- `user_can_mint_after_start_time` and `user_can_mint_before_end_time` are duplicates of `user_can_mint_during_mint_time`

We could probably delete the single ones and keep only the composite ones (`<xxx>can_mint_outside/during_mint_time`).

giorgionocera (commented):
Thank you again for your work.
Only one thing which could improve readability (it is not a bug or an improvement, it is only a suggestion, you can also discard it):
maybe you could move the `self.assert_user_dont_own_a_poap(deps.storage, &user_addr)?;` instruction inside the `mint_to_user` and avoid using it in the `mint_to` and `assert_user_can_mint` functions.
Feel free to do your evaluations.

manu0466:
> Overall looks good to me. However, it seems to me that there are some duplicated tests. Particularly:
> 
> * `admin_can_mint_before_start_time` and `admin_can_mint_after_end_time` are duplicates of `admin_can_mint_outside_mint_time`
> * `minter_can_mint_before_start_time` and `minter_can_mint_after_end_time` are duplicates of `minter_can_mint_outside_mint_time`
> * `user_can_mint_after_start_time` and `user_can_mint_before_end_time` are duplicates of `user_can_mint_during_mint_time`
> 
> We could probably delete the single ones and keep only the composite ones (`<xxx>can_mint_outside/during_mint_time`).

The tests are actualy different since in the `xxx_can_mint_during_mint_time` the contract is initialized with bot start and end time while in the `xxx_can_mint_after_start_time` and `xxx_can_mint_before_end_time`  the contract is initialized with only the start time or end time.

RiccardoM (approved):
Looks good to me overall

